PR #1, opened 2015-08-10
Add module event_registration_add_by_partner.

yajo:
Aiming for https://github.com/OCA/l10n-spain/issues/174, this module will make it easier to add partners from the database as event attendants.

Odoo's default behavior is to write manually the name of each attendant from the event form, and if you want to link it to a database partner, then you must go to the attendants subform and do it from there.

This module allows that behavior, but also allows adding the name from a database partner from the main form. It's easier and less confusing for end users.

### DIFF
--- a/event_registration_add_by_partner/README.rst
+++ b/event_registration_add_by_partner/README.rst
@@ -1,0 +1,50 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Add event registrations by partner
+==================================
+
+This module was written to extend the functionality of events to support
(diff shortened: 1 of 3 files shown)


rafaelbn:
@Yajo this functionality is already implemented in this module https://github.com/OCA/event/tree/8.0/partner_event. It also add more improvements. Could you please check it and see if is enough for you? If so we can close this PR. Thanks

yajo:
Yikes, I did not notice. I'll try, thanks.

yajo:
Ignore that last commit please.

yajo:
@rafaelbn You were right, thanks!

PR #2, opened 2015-08-10
[8.0][event_registration_seat_limit] Limit how many seats can each registration have

yajo:
Aiming for https://github.com/OCA/l10n-spain/issues/174.

This module was written to extend the functionality of events to support setting a maximum and a minimum of seats per registration.

If you have an event where you need to know the name of every attendant, or where attendants can come only with one companion, you can configure it.


pedrobaeza:
Please call the module _event_registration_seat_limit_

event_seats_per_registration/README.rst
```diff
@@ -0,0 +1,57 @@
...
+
+Usage
+=====
+
+To use this module, you need to:
+
+* Enter an event form.
+* Use the new fields in the *Registrations* tab.
+
+For further information, please visit:
```
pedrobaeza: Put new "Try me on runbot" button

event_seats_per_registration/__openerp__.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+# © 2015 Grupo ESOC Ingeniería de Servicios, S.L.U.
+
+{
+    "name": "Seats per registration in events",
+    "version": "8.0.3.1.0",
```
pedrobaeza: Why 8.0.3.1.0 as version? Couldn't be 8.0.1.0.0?

event_seats_per_registration/__openerp__.py
```diff
@@ -0,0 +1,21 @@
...
+    "category": "Tools",
+    "author": "Odoo Community Association (OCA)",
```
pedrobaeza: Add Grupo ESOC here

event_seats_per_registration/models.py
```diff
@@ -0,0 +1,81 @@
...
+from openerp import _, api, fields, models
+from . import exceptions
+
+
+class Event(models.Model):
+    """Events enhaced with participants per registration limits."""
+
+    _inherit = "event.event"
+
+    seats_per_registration_max = fields.Integer(
```
pedrobaeza: Use a shorter technical name: `max_seats`. It will ease any code logic. You can put in the label a longer description

event_seats_per_registration/models.py
```diff
@@ -0,0 +1,81 @@
...
+        "Maximum of participants per registration",
+        default=0,
+        help="Registrations with more than this number of participants will "
+             "be forbidden. Set 0 to ignore this setting.")
+
+    seats_per_registration_min = fields.Integer(
```
pedrobaeza: `min_registration_seats`

event_seats_per_registration/models.py
pedrobaeza: OK, I see there's another `seats_max` field, so it can confussing, but you can let it as `max_registration_seats` (at least you save 4 characters)

pedrobaeza:
The icon is unrelated to the current artwork, so I finally ended up making a PR against your branch with some of the improvements I suggest (icon, renaming and PEP8): https://github.com/grupoesoc/event/pull/1

Icon references:
- https://github.com/OCA/event/blob/8.0/event_project/static/description/icon.png
- https://github.com/OCA/website/blob/8.0/website_event_register_free_with_sale/static/description/icon.png
- https://github.com/OCA/website/blob/8.0/website_event_register_free/static/description/icon.png

pedrobaeza:
It would be good also if you rearrange your py files into models and splitting each model in one file, but I'm not going to block the PR by this.

rafaelbn:
Force rebuilding for testing. Please @Yajo could you review and update last comments of @pedrobaeza ? Thanks

event_seats_per_registration/__openerp__.py
yajo: It's because this module was already in another repository before https://github.com/OCA/event existed, and it had versions back there. If you don't care much, I'd prefer to leave it as is.

yajo:
All done.

pedrobaeza:
:+1:

rafaelbn:
@Yajo please review this module https://github.com/Antiun/antiun-odoo-addons/tree/8.0/website_event_register_limit it would be interesting to add this in front-end (website)

rafaelbn:
Tested in runbot :+1:

event_registration_seat_limit/exceptions.py
```diff
@@ -0,0 +1,61 @@
...
+        self.min = min
+        super(TooFewParticipants, self).__init__(value % min)
+
+
+class TooManyParticipants(SeatsPerRegistrationError):
+    def __init__(self,
+                 max,
+                 value=_("You cannot register more than %d participants.")):
+        self.max = max
+        super(TooManyParticipants, self).__init__(value % max)
```
dreispt: I'm puzzled at what's the benefit of having specific exceptions, but I won't pick on that.

event_registration_seat_limit/models.py
```diff
@@ -0,0 +1,81 @@
...
+    @api.constrains("registration_seats_max",
+                    "registration_seats_min",
+                    "seats_max")
+    def _check_seats_per_registration_limits(self):
+        """Ensure you are not setting an invalid maximum."""
+
+        if self.registration_seats_min < 1:
+            raise exceptions.NeedAtLeastOneParticipant()
+
+        if self.registration_seats_max != 0:
```
dreispt: Nitpick: more pythonic to use:  `if self.registration_seats_max:`

event_registration_seat_limit/models.py
```diff
@@ -0,0 +1,81 @@
...
+            if (self.registration_seats_max <
+                    self.registration_seats_min):
+                raise exceptions.MaxSmallerThanMin()
+
+        if self.seats_max != 0:
```
dreispt: `if self.seats_max:`

event_registration_seat_limit/models.py
```diff
@@ -0,0 +1,81 @@
...
+
+        if self.nb_register < self.event_id.registration_seats_min:
+            raise exceptions.TooFewParticipants(
+                self.event_id.registration_seats_min)
+
+    def _default_seats(self):
+        """Set the default number of participants per registration."""
+
+        # Normalize cases when there's no record
+        event_id = self.event_id.id or self.env.context.get("active_id", False)
```
dreispt: Are you sure this is needed? I may be wrong, but AFAIK upon save you will have a record to work with without having to depend on the context active_id.

event_registration_seat_limit/exceptions.py
pedrobaeza: Me too. This is not the first time I tell to @Yajo about that. It adds a lot of lines of code with little benefit (only a small filter for except statements, where most of the times it doesn't matter what is the exception). For putting error messages, you already have the Warning(v8)/UserError(v9) exception.

yajo: It depends if you prefer more solid and explicit code, or less code. [PEP 20](https://www.python.org/dev/peps/pep-0020/) says:

>    Explicit is better than implicit.
>     Complex is better than complicated.
>     Readability counts.
>     Errors should never pass silently.
>     Unless explicitly silenced.
>     In the face of ambiguity, refuse the temptation to guess.
>     Namespaces are one honking great idea -- let's do more of those!

But it also says:

>    Simple is better than complex.

IMHO explicit exceptions fit better in Python's zen.

If you review the tests, you will see that each test is expecting exactly the explicit exception it must raise. This makes it stronger, because if I just expect _any exception_ and there is a different error, future Travis and similars would fail silently, and we'd have to make in the future what I am doing now, but:

>    Now is better than never.

event_registration_seat_limit/models.py
yajo: It is, but I'll make it better.

yajo:
@rafaelbn What you ask for should be fixed by adding a view similar to:

```
<template id="event_description_full"
          inherit_id="website_event_sale.event_description_full">
    <xpath expr="//t[@t-as='nb']" position="attributes">
        <attribute name="t-foreach">range(
            ticket.event_id.registration_seats_min,
            (ticket.event_id.registration_seats_max or
             ticket.seats_available) + 1)
        </attribute>
    </xpath>
</template>
```

But that would make this module to depend on `website_event`, so IMHO it's better to make it a separate module, marked to autoinstall when this one and `website_event` are installed.

event_registration_seat_limit/exceptions.py
pedrobaeza: But this is not general Python, and there are already some filters/restrictions in the process layers of ORM. This new exceptions can bypass them because they don't use the except clauses, but an identification with `is` operator, causing this to fail sometimes.

dreispt: You are misinterpreting "explicit". Odoo Exceptions are already explicit, you just add them detail, and with that additional code. More code, more complexity. 
So IMHO it does not uphold the Zen.

yajo:
You told me exceptions were complex, so here you have simpler and more beautiful ones.

If you ask me to create a more error-prone test suite, my final answer is: **NO**. Here we are tired of buggy code, and we are determined to have rock-solid code. Any other solution you give me will always get the same answer if it implies worse tests.

If you tell me that I can achieve that same test strength by checking the error message, then you are asking me to replace in the test suite this:

```
        with self.assertRaises(e.PreviousRegistrationsFail):
            self._registrations.create({"nb_register": 10})
            self._event.registration_seats_min = 11
```

For this:

```
        try:
            self._registrations.with_context(lang="en_US").create({"nb_register": 10})
            self._event.with_context(lang="en_US").registration_seats_min = 11
        except ValidationError as error:
            self.assertEqual(
                error.value,
                "There are already registrations that don't fit in "
                "the new limits of participants per registration. "
                "Change them before setting the limits. "
                "The error was: {.value}".format(error))
```

And my answer is: are you kidding?

If you tell me that anybody will use `is` operator to control an exception, my answer is: Who would ever want to do that and why? Exceptions are handled with `try...except` blocks **always**, which is subclass-ready.

If you tell me that code is longer because of exceptions, my answer is: of course it's longer. We want strong code, and if adding security adds LOC, no problem. I could completely remove the test suite if all you care about is LOC, but you surely agree with me in that it would be a very stupid decision.

So if you have a real argument against explicit exceptions, please share it, but otherwise please let's avoid recursion about these things.

event_registration_seat_limit/exceptions.py
```diff
@@ -0,0 +1,47 @@
...
+class SeatsPerRegistrationValidationError(exceptions.ValidationError):
+    """Base class for this module's validation errors."""
+    def __init__(self, *args, **kwargs):
+        self._args, self._kwargs = args, kwargs
+        value = self._message()
+        super(SeatsPerRegistrationValidationError, self).__init__(value)
+
+    def _message(self):
+        """Format the message."""
+        return self.__doc__.format(*self._args, **self._kwargs)
```
dreispt, line 17: I can't help insisting that IMHO this is overengineered for no real benefit.

event_registration_seat_limit/models.py
```diff
@@ -0,0 +1,78 @@
...
+                    "registration_seats_min",
+                    "seats_max")
+    def _check_seats_per_registration_limits(self):
+        """Ensure you are not setting an invalid maximum."""
+        for s in self:
+            if s.registration_seats_min < 1:
+                raise exceptions.NeedAtLeastOneParticipant()
+
+            if s.registration_seats_max:
+                if (s.registration_seats_max <
```
dreispt: A single `if` would do the job better.

event_registration_seat_limit/models.py
```diff
@@ -0,0 +1,78 @@
...
+                        s.registration_seats_min):
+                    raise exceptions.MaxSmallerThanMin()
+
+            if s.seats_max:
+                if s.registration_seats_max > s.seats_max:
```
dreispt: Use a single `if`

event_registration_seat_limit/models.py
```diff
@@ -0,0 +1,78 @@
...
+                    raise exceptions.MaxPerRegisterBiggerThanMaxPerEvent()
+
+            try:
+                s.registration_ids._check_seats_per_registration_limits()
+            except exceptions.SeatsPerRegistrationError as error:
+                raise exceptions.PreviousRegistrationsFail(error)
```
dreispt, line 45: You catch an error to report another error?

dreispt:
Some non-blocking nitipicks, but LGTM.

pedrobaeza:
@Yajo, what @dreispt tries to say and I agree is that you don't need to design all for being so precise. In fact, if you have a test checking a raise, you'll have another with the flow without that raising condition, so it's not so important to know exactly what exception it is, so a general Warning/UserError is enough. If any other problem comes, you'll find it on the other test, so test suite can be the same "rock solid proof".

event_registration_seat_limit/exceptions.py
yajo, line 17: spam

event_registration_seat_limit/models.py
yajo, line 45: Error message from `SeatsPerRegistrationError` can be confusing when raised from here. `PreviousRegistrationsFail` makes that message easier to understand.

yajo:
@pedrobaeza This time I agree you are right. However I'm very interested in knowing the exact exception also because my company wants to have also fool-proof code. It means showing the correct error message and being sometimes a bit more precise among other things.

dreispt:
@yajo There is no additional effort to have fool-proof code and correct error messages without using custom exceptions. On the contrary, you have less code, arguably meaning less complexity and maintenance burden.

dreispt:
@yajo You have 2 OKs, but can you rebase? Strangely we don't see TravisCI result here, and running it there is fails for some reason. Maybe the rebase fixes that.

yajo:
@dreispt You cannot check the correct error message is raised without using custom exceptions or any of the ugly hacks explained in https://github.com/OCA/event/pull/2#issuecomment-163923575.

Rebased.

PR #3, opened 2015-08-12
[8.0][event_track_generate] Quick event tracks generator

yajo:
Aiming for https://github.com/OCA/l10n-spain/issues/174, this module generates event tracks automatically.

PR #13, opened 2015-11-05
Training modules

yajo:
This PR adds 3 modules aiming for https://github.com/OCA/l10n-spain/issues/174. This is the biggest PR of all that bug, and it creates a whole new "app" for Odoo: **Training**.
# Training

This module was written to extend the functionality of events to support
trainings and allow you to manage training actions, training groups,
training hour types, training materials, etc.
# Training Tracks

This module combines the power of modules `training` and
`website_event_track`.

Event tracks can get a training duration type, so it lets the user know if
the training group has fulfilled all the expected hours for it.
# Training track generator

This module was written to combine the functionality of the `training_track`
and `event_track_generate` modules and add the ability to set the training
hour type when batch generating event tracks in training groups.

For this reason, this PR is based on of https://github.com/OCA/event/pull/3 and includes module `event_track_generate`, or it would fail. You might want to merge that one before.


dreispt:
Can you split this in a separate PR for each module?

yajo:
I don't think so, because all git history is mixed, and all modules depend on the first one, so every PR should have to include it.

pedrobaeza:
You can use this technique (or directly use git filter-branch if you know that command): https://github.com/odoomrp/odoomrp-wip/wiki/Migration-to-OCA

yajo:
OK

dreispt:
The work in progress commit history is not relevant to keep.
In fact, is is good practice to squash it into only one or a few significant commits/changesets.

PR #8, opened 2015-09-29
[8.0][event_sale_pro] Better linking between events and their sale process

yajo:
Links registrations to other objects (sale.order and sale.order.line by default).

Also allows to generate quotations from the event registrations.

For more details see README.

Related to https://github.com/OCA/l10n-spain/issues/174.

PR #14, opened 2015-11-06
[8.0][training] Training module

yajo:
Relaunch first split module from https://github.com/OCA/event/pull/13. Target: https://github.com/OCA/l10n-spain/issues/174. CC @dreispt @pedrobaeza.
# Training

This module was written to extend the functionality of events to support
trainings and allow you to manage training actions, training groups,
training hour types, training materials, etc.


pedrobaeza:
Now what I'm going to ask you is to squash commit history to have less commits. Please make me happy :smile:

yajo:
I think this time I like it this way, sorry. :sweat_smile: This has been a hard work that started august 10th, and I like to have this proof of it as a personal :trophy:. Also many things have changed because many tests have been done, and if anybody asks _why is this the way it is?_, git blame could answer easily. I hope I don't make you sad :wink:

pedrobaeza:
I'm not talking about destroy all history, but commits like this https://github.com/grupoesoc/event/commit/53a84200e72a1ed9ec9d33c5b17e0fb3d1e4d1db can be squash with the main commit for that logical separation. 44 commits are too much for only 1 module.

training/__openerp__.py
```diff
@@ -0,0 +1,45 @@
+# -*- coding: utf-8 -*-
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+# © 2015 Grupo ESOC Ingeniería de Servicios, S.L.U. - Jairo Llopis
+
+{
+    "name": "Training",
+    "version": "8.0.3.0.0",
+    "category": "Project",
```
dreispt: I would expected "Event" here.

training/data/training.material.csv
```diff
@@ -0,0 +1,6 @@
+id,name,type_id:id
+material_pen,Pen,material_type_physical
+material_study_book,Study book,material_type_physical
+material_notebook,Notebook,material_type_physical
+material_username,User name,material_type_electronic
+material_password,Password,material_type_electronic
```
dreispt: IMO this looks is better as demo data, instead of default data.

training/models/event_event.py
```diff
@@ -0,0 +1,62 @@
...
+class Event(models.Model):
+    """Expand events with training actions.
+
+    Events with a training type and a training action are considered training
+    groups.
+    """
+    _inherit = "event.event"
+
+    material_ids = fields.Many2many(
+        M % "material",
```
dreispt: I find this unnecessary, and can't see a clear benefit for it.
What is the problem with just using `training.material`?
(this goes for all the other models)

training/models/training_material.py
```diff
@@ -0,0 +1,44 @@
...
+
+class MaterialType(models.Model):
+    """Type of the training materials."""
+    _name = M % "material_type"
+    _sql_constraints = [("unique_name",
+                         "UNIQUE(name)",
+                         "Name must be unique.")]
+
+    name = fields.Char(required=True, index=True, translate=True)
+    material_ids = fields.One2many(
```
dreispt: IMO this inverse relation is not really useful.

training/models/training_material.py
```diff
@@ -0,0 +1,44 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+# © 2015 Grupo ESOC Ingeniería de Servicios, S.L.U. - Jairo Llopis
+
+from openerp import fields, models
+from .common import M
+
+
+class MaterialType(models.Model):
+    """Type of the training materials."""
+    _name = M % "material_type"
```
dreispt: The common practice would be to name it `training.material.type`

training/models/training_material.py
```diff
@@ -0,0 +1,44 @@
...
+    material_ids = fields.One2many(
+        M % "material",
+        "type_id",
+        string="Materials",
+        help="Available materials of this type.")
+
+
+class Material(models.Model):
+    """Training material delivered to students."""
+    _name = M % "material"
```
dreispt: Did you consider using `product.product`?

training/models/training_action.py
```diff
@@ -0,0 +1,151 @@
...
+
+class ActionType(models.Model):
+    """Types of training actions.
+
+    Depending on the training action type, a training action may expect some
+    type of hours. For example:
+
+    - If a training action's type is "on-site", it expects on-site hours.
+    - If its type is "online", it expects online hours.
+    - If its type is "mixed", it expects both on-site and online hours.
```
dreispt: Reading this, I get the idea that the Type should be a computed field in the Training model, based on the values of "on-site" and "online" flags or tags.

training/models/training_action.py
```diff
@@ -0,0 +1,151 @@
...
+    expected_duration_type_ids = fields.Many2many(
+        M % "duration_type",
+        string="Expected hour types",
+        help="These types of hours are expected in this type of training "
+             "action. For example, a training of type 'mixed' may expect "
+             "hours of types 'on-site' and 'online'.")
+
+
+class Action(models.Model):
+    """Define training actions.
```
dreispt: I find this name confusing. AFAICS you mean "Courses"

training/models/training_action.py
```diff
@@ -0,0 +1,151 @@
...
+
+    name = fields.Char(required=True, index=True, translate=True)
+    type_id = fields.Many2one(
+        M % "action_type",
+        "Training type",
+        required=True)
+    contents = fields.Html(
+        translatable=True,
+        help="Contents of the course, shown in the rear side of the diploma.")
+    append_template = fields.Selection(
```
dreispt: `certificate_template` would be more appropriate

training/models/training_action.py
```diff
@@ -0,0 +1,151 @@
...
+    _name = M % "action"
+
+    name = fields.Char(required=True, index=True, translate=True)
+    type_id = fields.Many2one(
+        M % "action_type",
+        "Training type",
+        required=True)
+    contents = fields.Html(
+        translatable=True,
+        help="Contents of the course, shown in the rear side of the diploma.")
```
dreispt: IMO certificate is a better word than diploma

training/models/training_action.py
dreispt: `certificate_layout`?

training/models/training_action.py
```diff
@@ -0,0 +1,151 @@
...
+    duration_ids = fields.One2many(
+        M % "duration",
+        "action_id",
+        "Expected hours",
+        copy=True,
+        help="Expected duration of each type of hours for these training "
+             "groups.")
+    event_ids = fields.One2many("event.event", "training_action_id", "Events")
+    grade_min = fields.Float(
+        "Minimal grade",
```
dreispt: "Minimum"

training/models/training_action.py
```diff
@@ -0,0 +1,151 @@
...
+    def _check_grade_limits(self):
+        """Ensure no conflicts with grade limits."""
+        for s in self:
+            # Grade limits must be coherent
+            if not s.grade_min <= s.grade_pass <= s.grade_max:
+                msg = (_("Minimum grade cannot be bigger than maximum.")
+                       if s.grade_min > s.grade_max
+                       else _("Passing grade must be between minimum and "
+                              "maximum grades."))
+                raise ex.GradeLimitIncoherentError(msg)
```
dreispt: Why a specific Exception. IMO ValidationError would perfectly do the job.

training/models/training_action.py
```diff
@@ -0,0 +1,151 @@
...
+        "Training type",
+        required=True)
+    contents = fields.Html(
+        translatable=True,
+        help="Contents of the course, shown in the rear side of the diploma.")
+    append_template = fields.Selection(
+        ((2, "2 columns"), (3, "3 columns"), (4, "4 columns")),
+        help="Append one of these templates to the diploma contents. "
+             "They will help you to achieve some complex layouts.")
+    duration_ids = fields.One2many(
```
dreispt: Wouldn't it be simpler and more flexible to have the duration hours here, and avoid an extra duration codes table?

training/models/training_action.py
```diff
@@ -0,0 +1,151 @@
...
+        help="Students above than this grade will pass the training.")
+    grade_max = fields.Float(
+        "Maximum grade",
+        default=10,
+        required=True,
+        help="Students cannot get more than this grade.")
+    material_ids = fields.Many2many(
+        M % "material",
+        string="Materials",
+        help="These should be delivered to every student in this training.")
```
dreispt: I miss a "Course Category" field here.

training/models/event_event.py
```diff
@@ -0,0 +1,62 @@
...
+        string="Materials",
+        states={"done": [("readonly", True)]},
+        help="These should be delivered to every student in this training.")
+    training_action_id = fields.Many2one(
+        M % "action",
+        "Training action",
+        states={"done": [("readonly", True)]},
+        help="Training action of this event, if it is a training group.")
+    training_mode = fields.Boolean(
+        compute="_compute_training_mode",
```
dreispt, line 28: This is over-engineered. You just need a `is_tranining` boolean, that is set automatically to True with a `default_is_training: True` in the menu action context.

elicoidal:
@dreispt I made a bunch of comment on PR#16. Wasnot it the place? I am confused by all the PR

training/views/diploma_report.xml
```diff
@@ -0,0 +1,121 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+     © 2015 Grupo ESOC Ingeniería de Servicios, S.L.U. - Jairo Llopis -->
+
+<openerp>
+<data>
+
+<!-- Paper format for diplomas; landscape A4 with HiDIP -->
+<record id="diploma_paperformat" model="report.paperformat">
+    <field name="name">European Landscape A4 for Diplomas</field>
```
dreispt, line 10: How is this different from a A4 landscape paper format?

dreispt:
@yajo Thank you for this module, it is interesting to me and I would like to see it merged.
However it's quite large and I couldn't complete my review now.
At this point, my early opinion t I'm afraid that it can be over-engineered in some point (I believe simple is better) and I feel that it can be quite "opinionated" / too specific at some points, limited reuse for different use cases.
But we're here to discuss that and get an excellent result, so most of my comments are question, to help me get a better picture about your design choices.

dreispt:
IMO Courses should be Products.
You have the use case where you sell and facilitate trainings, but there is also the use case where you manage and buy trainings.
For the later, you want to know what suppliers provide a Course, and their catalogue price. You might also want to link to the procurement of courses.
This design might also be useful for companies selling training, since their catalogue would be in the Products table, and would fit nicely in other management areas, such as sales orders and invoices.

training/__openerp__.py
yajo: There is no _Event_ category. Event module goes into _Tools_, but that did not seem a good place for training. Not sure of what to do here...

training/models/event_event.py
yajo: DRY.

training/models/training_material.py
yajo: I'm more on the mood of the `module.model` schema, same as views. Seems more homogeneous to me. Maybe other community members can opine.

training/models/training_material.py
yajo: Actually it is a good idea, I did not think about it. But I think that complicates things... What should happen when a product is delivered? Maybe it's included in the ticket price, but maybe it's sold apart. How to handle that?

training/__openerp__.py
pedrobaeza: You can create the categories you need. I think this deserves a new category: "Event Management"

training/models/training_action.py
yajo: It could happen if `expected_duration_type_ids` were a Selection field, but being a Many2many what you say is not possible.

training/models/training_action.py
yajo: Yes, I put it like this to distinguish between what defines a course (currently training action) and what uses a course (currently training group), but being training groups events after all, your suggestion seems better. I'll change it.

training/models/training_material.py
dreispt: That goes against the common practice of using only "."

training/models/training_action.py
yajo: If you check the tests, you will see that they use all different exceptions in this module to check they are raised when needed. Without concrete exceptions, any other ValidationError would pass the test.

training/models/event_event.py
dreispt: DRY applies for code, not for repeating a word. I see disadvantages, when using grep for example.

training/models/training_action.py
yajo: Simpler yes, but not more flexible. Duration table lets the manager to set how many hours of each type are expected. Maybe a course has 30 on-site hours and 10 online hours. No way to get that without an extra table, sorry.

training/models/training_material.py
dreispt: You don't need to use any stock/logistics features. Just use the Product to describe courses.
You could setup some business logic regarding delivery, invoicing or purchasing - that's out of scope here, but the possibility is very interesting.

@pedrobaeza what do you think?

training/models/training_action.py
yajo: Yikes, true!

training/models/training_material.py
pedrobaeza: Yeah, it's a pity to not use product for this purpose. There's no problem with deliveries and so on. You can control your flow with for example invoice_control field to say if it's invoiceable or not.

training/data/training.material.csv
dreispt: I [found](https://github.com/odoo/odoo/blob/52484b0132ee3e0063dd67cc3c2eb434f99f5db8/openerp/addons/base/module/module_data.xml#L76) the correct name: use "Events"

training/models/training_action.py
dreispt: That's not consistent with the docstring.
The `expected_duration_type_id` relates to a `training.duration_type` model, but I can't find it.
I'm also unable to find the `training.duration` model.

dreispt:
@elicoidal The other PRs are repeating the code here, so that causes people reviewing the same code in separate PRs. I marked them as WIP to avoid reviews, but you were quicker than me :-)

training/__openerp__.py
yajo: OK

yajo: In such case I think I prefer just "Event" :wink:

dreispt: I [found](https://github.com/odoo/odoo/blob/52484b0132ee3e0063dd67cc3c2eb434f99f5db8/openerp/addons/base/module/module_data.xml#L76) the correct name: use "Events"

training/models/training_material.py
yajo: Then I'll just drop materials and use products, and leave further logic to be implemented in other modules as needs arise. Should I use product.product or product.template?

pedrobaeza: Better to put product.product, which is at the end the minimal unit for all (picking, invoices...).

training/models/event_event.py
yajo, line 28: It's not that simple. Users can use events from Marketing menu (as always) or from Training menu. Marketing allows you to convert an event that was a training group to a normal event, but the Training menu does not, because thanks to this trick the course field is required when using that menu.

For what you say, the logic is similar, but depends on _where was the record created_, not _where is the user right now_, which is much easier to understand, user POV.

yajo:
@elicoidal Yes, sorry for that. That's why I started with all together at #13, but I had to split it. I'll review your comments there too anyway, but I'll continue discussion here. Thanks for your help :wink:

training/views/diploma_report.xml
yajo, line 10: HiDPI, a must for diplomas.

elicoidal:
@Yajo @dreispt well noted

pedrobaeza:
I still insist that the commit history must be squashed. You can preserve the commit history in your own branch, but here in OCA it doesn't make sense to add 44 commits for one module.

dreispt:
@Yajo Can you comment on my suggestion to also make "Courses" a ´product.product`?

dreispt:
The commit history should the a change history, not an edit history. As is it brings little value, so I agree with Pedro that it should be squashed.

yajo:
Sorry I had no time to answer everything yesterday.

About squashing, fine. It sounds like "nobody cares you spent 3 months in this, we prefer to save 30KB of bandwidth", but nevermind. I'll do it as the last step because I will have to refactor many things for what I'm seeing, and it would grow again.

About course=product, I'm thinking on it. I made training a submodel of event, and since events get sold by ticket, there's where I thought the sale process should focus. But what you say makes sense. The problem is that it makes sense for events too.

I mean: being an event organizer company, you could sell a wedding, a congress, a concert, etc. You could even make variants by the expected size and sell it as a whole, not by tickets. This adds lots of flexibility to event sales, and training will still need to depend on that and add more things, but it blows this current implementation.

Also, to implement all that sale logic, we could use [event_sale_pro](https://github.com/OCA/event/pull/8) module, but that can be a latter extension to it.

So please mark this as WIP, let me do the `event_product` PR and refactor all this.

Thanks everyone!

dreispt:
Please share you design before you close the implementation: if you get early feedback on it you can avoid rework caused bu later changes to the design.

My functional PoV:

You already have from Events:
- "Training Events" (Groups in the module current terminology) map to Events. Organizing companies can sell tickets for them, just like in regular Events.
- Event Registrations are the training participants / attendees.
- Event Tracks are the training sessions

You need to add:
- "Training Courses" is a catalog of available courses. It can be implemented using delegation inheritance on Products, with some additional attributes specific to training. Each Training Event is related to a Training Course.
- "Training materials" is a many to may relation between Courses and Products.
- "Materials delivered" is a many to many relation between Registrations and Products.

Some details:
- The attributes added to Courses are important, we need to have some categorizations, and localizations will need to add legal reporting codes.
- The duration in hours statistics are very important. We need to know "contact time" (with a facilitator) vs "self study/assignments" time, and "labour time" and "outside labour time" hours. Probably the most flexible solution is to have a one 2 many in the Event model.
- Support for Evaluations is also a nice to have, but probably out of scope now.

yajo:
Well let me summarize then.
- I just squashed previous history and fixed all your remarks here and spell-checking in other PRs.
- I removed the `M %` trick that you did not like, it's true it allows for easier grepping.
- I put the right category.
- I removed all training materials stuff and replaced it by `product.product`. Amazingly, with almost no changes, all test suite works!

I felt that I needed to change all that to start redesigning, so now about design:

I like yours, @dreispt, and after some thinking, this is my plan:
1. Add more functionality to `event_sale_pro` (see #8):
   - Add a boolean to `product.template`: `is_event`.
   - Reuse `event_type_id` from `event_sale` for event products too.
   - Add a _product_ field to any event, allowing the user to have a catalog of events, and allowing to sell them as a whole instead of by registration.
   - Move there all logic found here about training materials. In events, it's quite common to deliver some present to attendees, so that can be controlled there.
     - Event products will have a new tab to set up the products to deliver.
2. Base this module on `event_sale_pro`, with some additions:
   - Add a boolean to `event.type`: `is_training`.
   - Event products where `event_type_id.is_training == True` will have all course stuff in their _event_ new tab.

An alternative idea is to remove all that and just add a _product_ field to `training.course`.

> The duration in hours statistics are very important. We need to know "contact time" (with a facilitator) vs "self study/assignments" time, and "labour time" and "outside labour time" hours. Probably the most flexible solution is to have a one 2 many in the Event model.

Can you explain yourself better? I think I do not fully understand this.

I'll thank feedback before starting implementation, please.

yajo:
I'd like to begin with implementation ASAP, please any comments on https://github.com/OCA/event/pull/14#issuecomment-158375281?

dreispt:
Unless I misunderstood your design is different from what I described.
I think we should have this:
- "Training Course"\* should be a new model with delegation inheritance on Products.
- "Event" is not a Product and has a Many2one field to the related "Training Course". When not empty it signals that this is a Training Event.

I'm not sure about the dependency on the other module. Why is it needed?
Could you write up an updated design? Maybe using a Google Doc would be easier to comment and edit.

yajo:
It's a bit different, yes, but not too much. Let me summarize our _idea diff_:
- Courses should be products.
  - All right, but then non-training events too.
- Delivered materials should be products.
  - All right, but those could be delivered to non-training event attendees too.

So I was proposing to implement that in a separate module, and base this one on that.

Since there is the open PR https://github.com/OCA/event/issues/8 (`event_sale_pro`), it seems a good place to put that, because if you make events and delivered materials to be products, it's because you want to sale them.

However right now `event_sale_pro` adds some functionality that might not seem related to this, so another approach would be putting all of it in some other module like `event_product`.

So the questions are:
- Is this design OK?
- Why use delegation inheritance instead of extending product model?
  - I could extend `product.template` and have event/course variants out of the box.
- Should I extend `event_sale_pro` or create a new `event_product` module?

dreispt:
> Is this design OK?

TBH it's not totally clear to me. The design discussion here ends up quite fragmented. Would you like to write it in a Google doc as I suggested?

> Why use delegation inheritance instead of extending product model?
>        I could extend product.template and have event/course variants out of the box.

OK, you won me with the product variants, that's an interesting option. 
A `is_event` would be OK.

> Should I extend event_sale_pro or create a new event_product module?

I dislike the dependency on event_sale_pro, since it forces me to install the heavyweight Sale module. Go for a event_product, but I would like to see it's design.

yajo:
> Would you like to write it in a Google doc as I suggested?

OK, I'm on it.

yajo:
There it is: https://docs.google.com/document/d/1WiOB1yAWTR2aROpM6O632QWXwgG4hLRUpTHU2Pv4yEQ/edit?usp=sharing

I'm not mentioning _training_track_ nor _training_track_generate_ because those are much simpler and I do not want to confuse you.

Please feel free to comment there.

yajo:
Please review new version of document.

I think I prefer the previous version where _event_sale_pro_ holds most things. This newer version needs 3 more modules to get exactly the same functionality, just for the sake of _not installing sale_. That is too much work for too little benefit IMHO.

This is the first version, in case you want to compare:

---

# Training module design

How to implement a training module for Odoo, and how all related modules will interact.

## event_product (or [event_sale_pro](https://github.com/OCA/event/pull/8))
- Purpose. 
  - Allow to sell events as a whole as products, not only tickets for them (that is what event_sale does). 
  - Allow to specify products that should be delivered to all event attendees. 
- Examples. 
  - You have this catalog of events, with variants: 
    - Wedding. 
      - Variants: 
        - Until 50 attendees. 
        - Until 100 attendees. 
        - In beach. 
        - In countryfield. 
      - Delivered products to attendees: 
        - Present. 
    - Congress. 
      - Variants: 
        - Until 50 attendees. 
        - Until 100 attendees. 
      - Delivered products to attendees: 
        - Present. 
    - Opening. 
      - Variants: 
        - 1 day. 
        - 2 days. 
        - 3 days. 
      - Delivered products to attendees: 
        - Leaflet. 
        - Pen. 
- Implementation. 
  - product.template: 
    - Add boolean is_event. 
      - Allows to have a catalog of events. 
      - Allows to sell events as a whole, not only by tickets. 
      - Most useful for event organizing companies. 
      - Disables field event_ok (cannot be a ticket and an event at the same time). 
    - Reuse event_type_id from event_sale for event products too. 
    - Add One2many delivered_product_ids: 
      - Products that should be delivered to attendees to that event. 
  - event.event: 
    - Add Many2one product. 
      - Points to product.product, where is_event == True and product.event_type_id in (event.event_type_id, False). 
    - Add One2many delivered_product_ids: 
      - Products that should be delivered to attendees to that event. 
      - Autofilled when choosing product, but customizable per event. 
    - Events linked to a Product with the "is_event" flag will have a new tab to set up the products to deliver. 
  - event.registration: 
    - Add boolean products_delivered. 
    - Add product delivery receipt report. 
- Ideas for future versions: 
  - Product delivery tracking. 
  - Statics. 
  - Create event on sale of product. 

## [training](https://github.com/OCA/event/pull/14)
- Purpose. 
  - Extend event with things needed for managing training sessions. 
    - Print diploma/certificate for attendees. 
    - Grade. 
    - Courses. 
      - Types. 
    - Duration. 
      - Types. 
- Implementation: 
  - Based on event_product. 
  - training.duration_type: 
    - Defines a type of training hours (on-site, online…) 
  - event.type: 
    - Add boolean is_training. 
  - product.template: 
    - If event_type_id.is_training: 
      - This product is a course. 
      - HTML field contents. 
        - Customize back side of diploma/certificate. 
        - Has built-in template generator. 
      - One2many duration_ids (new training.duration model). 
        - Autofilled from event_type_id.expected_duration_type_ids. 
        - Defines the amount of hours of each type expected for these events. 
      - Float grade_min. 
      - Float grade_pass. 
      - Float grade_max. 
      - Add Many2many expected_duration_type_ids. 
        - What type of hours are expected in this type of training. 
  - event.registration: 
    - Float grade. 
    - Computed float passing. 
      - Depends on grade and product.grade_pass\* fields.

dreispt:
@Yajo Google Docs already allows to see the changes.
I made a diagram of the dependencies, here: https://drive.google.com/file/d/0B72Vums-E7ooSkg2UEFndzI4dDg/view?usp=sharing

My conclusion is that we need three modules: event_product, event_product_logistics, and event_training.

yajo:
And all of them must rely on training_sale, right?

dreispt:
What is the point for training_sale?

dreispt:
Selling Training is just selling Events. Unless I'm missing something, the `event_sale_pro` should handle it.

yajo:
Sorry I meant event_sale.

yajo:
I feel like we are walking in circles to avoid the unavoidable sale module. It is unavoidable because:
- `event_sale` adds `event_type_id` to `product.template`.
- We need that field for `event_product`.
- `event_sale` installs `sale`.
- `event_training` requires `event_product`.

The same problem exists with event tracks: you cannot install them without `website_event`, not because you actually need website for tracks, but because of a poor design decision from core.

I need to start implementation now, so I'll do it this way:
- Leave event_sale_pro as it is, seems fine and unrelated enough.
- Add event_product, which will include all related to selling events as products.
- Add event_product_deliverable, which will handle the delivery of products to attendees. I still see no problem with the name event_product_delivery because future versions should connect with warehouse, but I don't want to fight anymore for 4 characters.
- Add event_training, which will include whatever needed to create training products.

I'm closing this and will open new PR for each module, this one is too long already.

yajo:
Design doc updated FWIW.

rafaelbn:
I agree :+1: 
- event_product
- event_product_logistics
- event_training

Why event_product_sale? We need event type in event_product and this field is from event_sale, we must use the same.

https://github.com/OCA/OCB/blob/8.0/addons/event_sale/event_sale.py#L32
- event_sale_pro is not related with this discussion, in fact event_sale_pro must change name as pro is not clear and the main purpose of the module is to give the functionality of creating sale order from events

dreispt:
@Yajo I really want to skip the sale dependency.
But I don't want to block your work.
Go ahead and I'll later try a PR to remove the dependency form `event_product`.

By the way, I tried to make made the design diagram collaborative. 
Follow the link and pick the draw.io app to view/edit:
https://drive.google.com/file/d/0B72Vums-E7oodlplWlhMQ2MxNVk/view?usp=sharing

dreispt:
Note: I have a WIP on design suggestions to event_training...

yajo:
@dreispt I saw your suggestions, but those go to event_training_track, which is much simpler and will come later. You can see the previous #15, but it's outdated now. Again, that will make training depend on website because Odoo's mania of merging too much functionality in the same module. But let that stuff for its moment.

dreispt:
OK, let's get back to that at that point.
Please share a design before you start working on that area.

PR #15, opened 2015-11-06
[8.0][training_track] Adapt event tracks to trainings

yajo:
Relaunch second split module from https://github.com/OCA/event/pull/13. Merged with #14 to make it work. Target: https://github.com/OCA/l10n-spain/issues/174. CC @dreispt @pedrobaeza.
# Training Tracks

This module combines the power of modules `training` and
`website_event_track`.

Event tracks can get a training duration type, so it lets the user know if
the training group has fulfilled all the expected hours for it.


training/README.rst
```diff
@@ -0,0 +1,163 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+========
+Training
+========
+
+This module was written to extend the functionality of events to support
+trainings and allow you to manage training actions, training groups,
```
elicoidal, line 9: training is no plural (check PR#16)

dreispt:
@yajo One module per PR please. Isn't `training` already in #14? This PR will have conflicts when that is merged. Just signal in the description that it depends on #14.

yajo:
This seems to confuse reviewers, and #14 is being totally refactored so this will have to change too. Closing until #14 is merged.

PR #22, opened 2015-12-02
[FIX] Bufixes in event to project sync

antespi:
Fixes:
- Set `project.event_id` to `False` when removing project from event
- Set `project.event_id` properly when creating project from template
- Set a restricted domain to `event.project_id` field in order to show only live projects without a event associated


pedrobaeza:
:+1:

rafaelbn:
:+1:

JavierIniesta:
:+1:

PR #21, opened 2015-11-27
[8.0][event_product] New module to combine events and products

yajo:
Following discussion from #14, and aiming for https://github.com/OCA/l10n-spain/issues/174, this module links events and products.

This implementation follows [what was discussed before](https://docs.google.com/document/d/1WiOB1yAWTR2aROpM6O632QWXwgG4hLRUpTHU2Pv4yEQ/edit?usp=sharing), and of course this is the first of a collection of modules to implement training in Odoo.

See README for instructions, please.

CC @pedrobaeza @elicoidal @dreispt @rafaelbn.

Thanks!

PR #23, opened 2015-12-29
[8.0][WIP][event_product_logistics] Manage product delivery for event attendees

yajo:
As asked by @dreispt in https://github.com/OCA/event/pull/21#issuecomment-160631527, I'm opening this PR. Bots will fail until https://github.com/OCA/event/pull/21 is merged. Target: https://github.com/OCA/l10n-spain/issues/174. [Design](https://docs.google.com/document/d/1WiOB1yAWTR2aROpM6O632QWXwgG4hLRUpTHU2Pv4yEQ/edit?usp=sharing) as discussed in #14.

Sometimes you want to give some products to your event's attendees. For example: a pen, a notebook, a book, etc.

This module extends the functionality of events to support setting products to deliver to event attendees and allow you to know if they have been already delivered to them.


event_product_logistics/README.rst
```diff
@@ -0,0 +1,120 @@
...
+
+========================================
+Product Delivery for Event Registrations
+========================================
+
+Sometimes you want to give some products to your event's attendees. For
+example: a pen, a notebook, a book, etc.
+
+This module extends the functionality of events to support setting products to
+deliver to event attendees and allow you to know if they have been already
```
elicoidal: have already been

event_product_logistics/README.rst
```diff
@@ -0,0 +1,120 @@
...
+delivered to them.
+
+.. warning::
+    Please do not get confused between *event products* and *products to
```
elicoidal: make no confusion between

event_product_logistics/README.rst
```diff
@@ -0,0 +1,120 @@
...
+    deliver in events*:
+
+    * **Event products** are created in the ``event_product`` addon, and
+      define an event as a product, to have a catalog of them.
+    * **Products to deliver in events** are created in this addon, and define
+      the products that you will have to deliver to event attendees.
```
elicoidal: the products (goodies, leaflets, etc.) that you will deliver

event_product_logistics/README.rst
```diff
@@ -0,0 +1,120 @@
...
+This module extends the functionality of events to support setting products to
+deliver to event attendees and allow you to know if they have been already
+delivered to them.
+
+.. warning::
+    Please do not get confused between *event products* and *products to
+    deliver in events*:
+
+    * **Event products** are created in the ``event_product`` addon, and
+      define an event as a product, to have a catalog of them.
```
elicoidal: product (wedding, fair, etc.)

event_product_logistics/README.rst
```diff
@@ -0,0 +1,120 @@
...
+      They can be defined in the *event product*, in the event directly, or in
+      both.
+
+Usage
+=====
+
+To set products to deliver in an event product (see ``event_product`` module),
+you need to:
+
+- Go to *Sales > Products > Product variants > Create*.
```
elicoidal: better enumerated list for procedures

event_product_logistics/README.rst
```diff
@@ -0,0 +1,120 @@
...
+- Set a name.
+- Enable *Is an event*.
+- Open the new tab called *Event*.
+- Set some products to deliver there.
```
elicoidal: remove "some"

event_product_logistics/README.rst
```diff
@@ -0,0 +1,120 @@
...
+
+To set products to deliver in an event, you need to:
+
+- Go to *Marketing > Events > Events > Create*.
+- Set a name and start and end dates.
```
elicoidal: - Set a name, start and end dates.

event_product_logistics/README.rst
```diff
@@ -0,0 +1,120 @@
...
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/199/8.0
+
+Known issues / Roadmap
+======================
+
+* Please read known issues section of ``event_product``'s README about
+  installation of the ``sale`` addon.
+* Product deliveries should create the corresponding inventory movements.
```
elicoidal: Product deliveries currently do not create the corresponding stock moves.

event_product_logistics/__openerp__.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+# © 2015 Grupo ESOC Ingeniería de Servicios, S.L.U. - Jairo Llopis
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Product Delivery for Event Registrations",
+    "summary": "Deliver products to event attendees",
```
elicoidal, line 7: Usually name should be shorter than summary ;)

event_product_logistics/__openerp__.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+# © 2015 Grupo ESOC Ingeniería de Servicios, S.L.U. - Jairo Llopis
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Product Delivery in Events",
+    "summary": "Deliver products to event attendees",
+    "version": "8.0.1.0.0",
+    "category": "Events",
```
dreispt: I'm not sure if this shouldn0t be "Event Management". Can you check?

event_product_logistics/models/product.py
```diff
@@ -0,0 +1,16 @@
...
+# © 2015 Grupo ESOC Ingeniería de Servicios, S.L.U. - Jairo Llopis
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    event_deliverable_product_ids = fields.Many2many(
```
dreispt, line 11: Why don't you add this to `event.event` instead.
With that you could have this module depend directly on `event` instead of `event_product`

event_product_logistics/views/delivery_receipt_report.xml
```diff
@@ -0,0 +1,71 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+     © 2015 Grupo ESOC Ingeniería de Servicios, S.L.U. - Jairo Llopis -->
+
+<openerp>
+<data>
```
dreispt, line 6: add noupdate=1 for easier report customization (or maybe that's not a worry on templates?)

event_registration_deliverable_product/README.rst
```diff
@@ -0,0 +1,100 @@
...
+    Please do not get confused between *event products* and *products to
+    deliver in events*:
+
+    * **Event products** are created in the ``event_product`` addon, and
+      define an event as a product, to have a catalog of them.
+    * **Products to deliver in events** are created in this addon, and define
+      the products that you will have to deliver to event attendees.
+
+      They can be defined in the *event product*, in the event directly, or in
+      both.
```
dreispt: Is it me or this reads the same as in the `event_product_logistics`? 
I then get the idea that both modules do the same.

dreispt:
I have two main design questions:
- Why depend on `event_product`? AFAICS that could be avoided.
- Why the two modules? The second one repeats code from the first. Maybe it could depend on the first to avoid such repetition.

event_product_logistics/views/delivery_receipt_report.xml
yajo, line 6: Yes, it does not affect templates. When you manually edit them, its XMLID's `noupdate` becomes `True` and no more updates are applied.

dreispt, line 6: ok

yajo:
Yikes, I forgot to remove the old module! Thanks!

The relation between events and products in this module can be seen in method `action_load_deliverable_products`, where you can load that information from the linked product.

Indeed, I could just create 2 modules instead of one, but the second would be so small that I decided to avoid that for simplicity, but if you feel it's necessary I can split it.

event_product_logistics/models/product.py
yajo, line 11: It's there too, please check.

dreispt:
No need to split.
I have the feeling that the `event_product` dependency can be avoided. What do you think?

yajo:
Normally, if e.g. your event is from product _Odoo onsite training_, you know you will give away _notebook, pen and Odoo t-shirt_. If you cannot define that in the product, you need to choose them manually in every event.

With current implementation, you give the user the chance to edit the deliverable products in the event, but to do it in the product to avoid repetition too.

How do you think we can get this without depending on `event_product`?

dreispt:
Instead of adding `event_deliverable_product_ids` to `product.template`, add it directly to `event.event`.
AFAICS all the rest should work.

yajo:
[It already has it](https://github.com/OCA/event/pull/23/files#diff-562588a7749fe3d7ae66ff937eef98fdR11).

dreispt:
So the same field is in the two models? 
I'm confused; what's the point?

yajo:
I think it will be clearer if you try it. Please open the runbot and follow all the steps in the usage section of the readme. Then please tell me what you don't understand.

dreispt:
OK, will try that.

dreispt:
I see the point.
In theory we could split the `product`-side features into another module, but as you say, in practice it would probably be an overkill. So OK: :+1:

PR #24, opened 2016-01-04
[8.0][event_training] Training module

yajo:
This module extends events to support training events and courses.
- Source: Refactorization of #14
- Design: [Google Docs](https://docs.google.com/document/d/1WiOB1yAWTR2aROpM6O632QWXwgG4hLRUpTHU2Pv4yEQ/edit?usp=sharing)
- Target: https://github.com/OCA/l10n-spain/issues/174
- Dependencies: #21 (bots will fail until merged)

Rebased, squashed all except one real commit that I hope will help to prevent the exception flamewar. Please avoid it in this thread, I'm tired of it.

PR #27, opened 2016-01-13
[8.0][WIP][event_training_track] New module to mange training tracks.

yajo:
- Source: Refactorization of https://github.com/OCA/event/pull/15.
- Design: [Google Docs](https://docs.google.com/document/d/1WiOB1yAWTR2aROpM6O632QWXwgG4hLRUpTHU2Pv4yEQ).
- Target: https://github.com/OCA/l10n-spain/pull/174.
- Dependencies: https://github.com/OCA/event/pull/21, #24 (bots will fail until merged).

This module takes `event_training` as a base and combines it with
`website_event_track`.

The result is that you can choose the expected duration for each training
event type and product. Events based on that product will be expected to
fulfill that amount of hours of each type. You can also check if those
expectations are fulfilled with a built-in wizard.


event_training_track/__openerp__.py
```diff
@@ -0,0 +1,37 @@
...
+    "author": "Grupo ESOC Ingeniería de Servicios, "
+              "Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "website": "https://grupoesoc.es",
+    "installable": True,
+    "application": False,
+    "auto_install": True,
+    "depends": [
+        "event_training",
+        "website_event_track",
```
dreispt, line 19: Can we please avoid the website dependency?

yajo, line 19: I'm afraid not, it's the module that defines the `event.track` model.

dreispt, line 19: I'm afraid you're right. Bad design choice from Odoo.

yajo, line 19: Yes, yet again, like with `event_sale` :confused:

PR #31, opened 2016-02-03
[8.0][ADD][WIP][event_multi_image] Image gallery for events.

yajo:
This module allows to have multiple images in each event.

It depends on:
- [x] https://github.com/OCA/server-tools/pull/374 to work.

@rafaelbn


dreispt:
Question: isn't there a product multi-image module? With the feature linking Events to Products there would be no need to have and event-specific multi-image.

yajo:
The event product could be i.e. "Conference", but each conference has its own gallery.

dreispt:
OK

.travis.yml
```diff
@@ -34,6 +34,9 @@ install:
   - git clone https://github.com/OCA/maintainer-quality-tools.git ${HOME}/maintainer-quality-tools --depth=1
   - export PATH=${HOME}/maintainer-quality-tools/travis:${PATH}
   - travis_install_nightly
+  - pip install python-ldap
+  - pip install unidecode
+  - pip install validate_email
```
dreispt: Why the new dependencies?

event_multi_image/models/event.py
```diff
@@ -0,0 +1,10 @@
+# -*- coding: utf-8 -*-
+# © 2016 Antiun Ingeniería S.L. - Jairo Llopis
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import models
+
+
+class EventEvent(models.Model):
+    _name = "event.event"
+    _inherit = [_name, "multi_image_base.owner"]
```
dreispt: Does it need to depend on itself?

dreispt:
Other than the added dependencies, LGTM.

.travis.yml
yajo: For product-attribute repo.

event_multi_image/models/event.py
yajo: It's doing it. `_name == "event.event"`

.travis.yml
dreispt: OK. With the new support for requirements.txt on repos this will be unnecessary once the requirements file is added to product-attribute.

event_multi_image/models/event.py
dreispt: I can see that, but why?

.travis.yml
yajo: Do you mean that I can remove this now, or that I will can in some near future?

event_multi_image/models/event.py
yajo: Why not?

yajo: Oh sorry I misunderstood your question :sweat_smile:. The point is that AFAIK to extend an already-existing module with some another you need to declare it both in `_name` and in `_inherit`. You can see the example in [crm_mass_mailing](https://github.com/odoo/odoo/blob/8.0/addons/crm_mass_mailing/mass_mailing.py#L6).

oca_dependencies.txt
```diff
@@ -1,4 +1,5 @@
 # List the OCA project dependencies, one per line
 # Add a repository url and branch if you need a forked version
+server-tools
```
yajo, line 3: Runbot says:

```
except_orm: ('Error', u"You try to install module 'event_multi_image' that depends on module 'base_multi_image'.\nBut the latter module is not available in your system.")
```

That module is in `server-tools` now. Is it not enough to add it here?

pedrobaeza, line 3: No, this is for Travis, but runbot isn't still able to process this. I'm adding it manually

dreispt:
LGTM :+1:

pedrobaeza:
:+1:

PR #32, opened 2016-02-04
[ADD] crm_lead_to_event_registration

pedrobaeza:
# 
# Register a lead directly in an event

This module allows to register a lead directly in an event when converting
this lead into a partner, so that you can save several steps.
# Usage

Go to any lead in _Sales > Sales > Leads_, and click on the button "Convert
to Opportunity".

In the resulting screen, if the conversion action is "Convert to opportunity"
and you have selected a related partner (a new or an existing one), you will
have a section called "Event registration".

In that section, you can select an event, and when you click on
"Create opportunity", the partner will be registered automatically in the
event.


pedrobaeza:
cc @rafaelbn

yajo:
I create the lead:

captura de pantalla de 2016-02-05 13-18-11

Then I do the steps, and see the registration:

captura de pantalla de 2016-02-05 13-18-45

No name, phone nor email.

Besides, nothing in the opportunity tells me it was already registered in an event. Is this by design?

And... No Coveralls here? :confused:

rafaelbn:
Hi, tested.

In the case you don't want to register a lead to an event but you want to register a opportunity to an event is missing a button (this button changes to view registration in the case is already done)



As @Yajo says, some data from partner is missing in the registration. Also needed the option of navigate from the opportunity to the registration



Thanks!

pedrobaeza:
You can't do it in the opportunity because there's no button to convert to partner. This will be a totally new wizard. About the button, I prefer a smart-button, but I don't like to fill the screen with shortcuts to every object in the system.

yajo:
You can use the "more" menu

rafaelbn:
functionally tested :+1:  just fix travis please

crm_lead_to_event_registration/wizard/crm_lead_to_opportunity.py
```diff
@@ -0,0 +1,25 @@
...
+    event_id = fields.Many2one(
+        comodel_name="event.event", string="Create registration in this event")
+
+    @api.multi
+    def action_apply(self):
+        res = super(CrmLead2OpportunityPartner, self).action_apply()
+        self.ensure_one()
+        if not self.event_id:
+            return res
+        registration_model = self.env['event.registration']
```
JavierIniesta: Travis:
crm_lead_to_event_registration/wizard/crm_lead_to_opportunity.py:21:9: F841 local variable 'registration_model' is assigned to but never used

crm_lead_to_event_registration/models/crm_lead.py
```diff
@@ -0,0 +1,43 @@
...
+
+    @api.multi
+    @api.constrains("probability", "stage_id")
+    def _check_status_confirm_registration(self):
+        """If the opportunity is won/lost, open/cancel registration."""
+        won = self._track["stage_id"]["crm.mt_lead_won"]
+        lost = self._track["stage_id"]["crm.mt_lead_lost"]
+        for s in self:
+            if won(s, s.env.cr, s.env.uid, s, s.env.context):
+                s.event_registration_id.registration_open()
```
antespi, line 46: Why do you execute this action if this is a constrain?. Constrain was designed to raise ValidationError when some conditions are triggered and return True if data is valid.
I'm not sure what could happen if any ValidationError is raised after action execute

IMHO, this action could be in write method

crm_lead_to_event_registration/wizard/crm_lead_event_pick.py
```diff
@@ -0,0 +1,24 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import api, fields, models
+
+
+class WizardModel(models.TransientModel):
+    _name = "crm.lead.event.pick"
+
+    lead_id = fields.Many2one(
+        "crm.lead",
```
antespi, line 12: Could you use explicit keyword parameters, as defined in guidelines? https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#field

```
    lead_id = fields.Many2one(
        comodel_name="crm.lead",
        string="Lead/Opportunity",
(...)
```

crm_lead_to_event_registration/wizard/crm_lead_event_pick.py
```diff
@@ -0,0 +1,24 @@
...
+        "Lead/Opportunity",
+        default=lambda self: self.env.context["active_id"],
```
antespi: Could you use a method for default? As described in guidelines:

```
Default functions should be declared with a lambda call on self. The reason for this is so a default function can be inherited. Assigning a function pointer directly to the default parameter does not allow for inheritance.

a_field(..., default=lambda self: self._default_get())
```

crm_lead_to_event_registration/wizard/crm_lead_event_pick.py
```diff
@@ -0,0 +1,24 @@
...
+        help="Lead/Opportunity from where to get contact information.")
+    event_id = fields.Many2one(
+        "event.event",
+        "Event",
+        required=True,
+        help="Event where the registration will be created.")
+
+    @api.multi
+    def action_accept(self):
+        self.lead_id.action_generate_event_registration(self.event_id)
```
antespi, line 28: Should this method support multirecord? (because of api.multi):

```
@api.multi
def action_accept(self):
    for wizard in self:
        wizard.lead_id.action_generate_event_registration(wizard.event_id)
```

or use `self.ensure_one()`

crm_lead_to_event_registration/__openerp__.py
```diff
@@ -0,0 +1,26 @@
...
+        'event',
+        'marketing_crm',
+    ],
+    "data": [
+        'wizard/crm_lead_event_pick_view.xml',
+        'wizard/crm_lead_to_opportunity_view.xml',
+        'views/crm_lead_view.xml',
+    ],
+    "installable": True,
+}
```
antespi: images key is not declared, Travis saids

crm_lead_to_event_registration/models/crm_lead.py
```diff
@@ -0,0 +1,43 @@
...
+    @api.multi
+    def action_generate_event_registration(self, event):
+        """Generate an event registration."""
+        er = self.env["event.registration"]
+        for s in self:
+            s.event_registration_id = er.create({
+                "event_id": event.id,
+                "partner_id": s.partner_id.id,
+                "nb_register": 1,
+                "name": s.contact_name,
```
antespi: Registration could fallback to `s.partner_name` and then `s.partner_id.name`. If none defined a ValidationError could be raised because in that case registration name would be 'False' as you can see in the snapshot

image

crm_lead_to_event_registration/models/crm_lead.py
yajo, line 46: Constraints are a trigger after all, that gets hit when a field is changed. That's exactly the behavior I want. Overwriting `write()` would cause an overhead in every `crm.lead` update and would mean a bunch of extra code just to get the same result.

crm_lead_to_event_registration/wizard/crm_lead_event_pick.py
yajo, line 12: :open_mouth: Is that seriously a guideline? I can't believe it, I thought it's just an example of model order declaration...

crm_lead_to_event_registration/wizard/crm_lead_event_pick.py
yajo, line 28: No need for multi in wizards, and when accessing a field from a record in attribute form (`self.lead_id`), `ensure_one()` is implicit.

crm_lead_to_event_registration/models/crm_lead.py
yajo: How do you get to that? I can't reproduce.

antespi: 1. Create a lead
2. Set lead name only (Company Name, Customer and Contact name empty)
3. Click 'Generate registration'

crm_lead_to_event_registration/wizard/crm_lead_event_pick.py
antespi, line 12: This is a good practice BTW because parameters order could change in future (not in Odoo, of course). If you set by keywords, changing order will not affect to this call.

crm_lead_to_event_registration/models/crm_lead.py
guewen, line 46: This is not something I would encourage to use. It has been discussed and here specifically an answer of @odony which is worth to read: https://github.com/odoo/odoo/issues/3518#issuecomment-62722436

simahawk, line 46: I agree w/ @guewen. If you don't want to add overhead on write (or whichever base method) just set a flag in the context and perform the action only if that flag is present.

antespi, line 46: > Odony: The behavior described in the issue really looks like an abuse of @constrains to me, and I'm not sure many developers would naturally go that route

@Yajo Could you change this and use a write pattern:

```
@api.multi
def write(self, vals):
    result = super(CrmLead, self).write(vals)
    self._check_status_confirm_registration(vals)
    return result

@api.multi
def _check_status_confirm_registration(self, vals):
   if any([x in vals for x in ("probability", "stage_id")]):
       # your code here, included a loop over self
```

crm_lead_to_event_registration/models/crm_lead.py
yajo: In such case, I think getting an empty name is the expected behavior, isn't it?

crm_lead_to_event_registration/wizard/crm_lead_event_pick.py
yajo, line 12: I hope they don't do that, but if they do, then it will be the time to fix this :wink:

crm_lead_to_event_registration/__openerp__.py
Yajo: Fixed.

crm_lead_to_event_registration/models/crm_lead.py
yajo, line 46: Man, this is bad news. Along with the deprecation of `@api.one` they keep removing features to the new api that looked so fine at the beginning. Let me change it.

crm_lead_to_event_registration/models/crm_lead.py
antespi: But this behavior has no sense. I think that implement the fallback and raise a ValidationError helps user and increase UX. And It's very easy to implement BTW

crm_lead_to_event_registration/models/crm_lead.py
antespi, line 46: I think that these features was (and some are) not well explained, for instance api.depends and api.onchange

yajo, line 46: You can see the current code. It's much uglier, and previous works. If they allow write access in constraints, they should not get surprised if developers use it, or they should provide a sane alternative. But most probably they don't care much.

guewen, line 46: As a matter of readability, if I read `api.constrains`, I expect it to be a constraint, the last thing I would expect is it to have side effects. I agree that an alternative for this use case would be nice though.

yajo, line 46: Anybody interested, please see https://github.com/odoo/odoo/pull/11042

pedrobaeza:
cc @Tecnativa

rafaelbn:
Please @pedrobaeza review comments. Rebuilding for testing. Thanks!

yajo:
I think the only remainder is https://github.com/OCA/event/pull/32#discussion-diff-53337467 to merge

pedrobaeza:
@Yajo, all is right, so please make a last check before merging.

crm_lead_to_event_registration/models/crm_lead.py
yajo: FTR the issue in this conversation is fixed now.

yajo:
I guess runbot's :x: is due to the failure that is happening lately, so :+1:

PR #36, opened 2016-02-17
Make no type count as a type.

### LINKED ISSUE #35: [8.0][BUG] event_registration_cancel_reason - You cannot cancel registrations from events of different types at once.
module: event_registration_cancel_reason
version: 8.0

**Steps to reproduce**
- Try to cancel a registration in an Event which doesn't have set "Type of Event"

**Current behavior**
- You cannot cancel registrations from events of different types at once. 

**Expected behavior**
- If there is a cancel reason for all Type of Events, then you must be able to cancel registrations of events which do not have any Type assigned.

@yajo @pedrobaeza

yajo:
Fix #35. When an event had no type, it was not getting in the mapping because of the way Odoo generates recordsets.

@rafaelbn @pedrobaeza

### DIFF
--- a/event_registration_cancel_reason/wizard/event_registration_cancel_log_reason.py
+++ b/event_registration_cancel_reason/wizard/event_registration_cancel_log_reason.py
@@ -23,7 +23,7 @@ class EventRegistrationCancelLogReason(models.TransientModel):
             var_fields)
         registrations = self.env['event.registration'].browse(
             self.env.context['active_ids'])
-        first_type = registrations[0].event_id.type
+        first_type = registrations[:1].event_id.type
         for event in registrations.mapped("event_id"):
             if event.type != first_type:
                 raise exceptions.ValidationError(
--- a/event_registration_cancel_reason/wizard/event_registration_cancel_log_reason.py
+++ b/event_registration_cancel_reason/wizard/event_registration_cancel_log_reason.py
@@ -23,12 +23,13 @@ class EventRegistrationCancelLogReason(models.TransientModel):
             var_fields)
         registrations = self.env['event.registration'].browse(
             self.env.context['active_ids'])
-        event_type = registrations.mapped("event_id.type")
-        if len(event_type) != 1:
-            raise exceptions.ValidationError(
-                _("You cannot cancel registrations from events of different "
-                  "types at once."))
-        res['event_type_id'] = event_type.id
+        first_type = registrations[0].event_id.type
+        for event in registrations.mapped("event_id"):
+            if event.type != first_type:
+                raise exceptions.ValidationError(
+                    _("You cannot cancel registrations from events of "
+                      "different types at once."))
+        res['event_type_id'] = first_type.id
         return res
 
     @api.multi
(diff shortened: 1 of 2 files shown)


event_registration_cancel_reason/wizard/event_registration_cancel_log_reason.py
```diff
@@ -23,12 +23,13 @@ def default_get(self, var_fields):
             var_fields)
         registrations = self.env['event.registration'].browse(
             self.env.context['active_ids'])
-        event_type = registrations.mapped("event_id.type")
-        if len(event_type) != 1:
-            raise exceptions.ValidationError(
-                _("You cannot cancel registrations from events of different "
-                  "types at once."))
-        res['event_type_id'] = event_type.id
+        first_type = registrations[0].event_id.type
```
pedrobaeza: Put `registrations[:1]`to avoid errors in case of 0 registrations.

event_registration_cancel_reason/wizard/event_registration_cancel_log_reason.py
```diff
@@ -23,12 +23,13 @@ def default_get(self, var_fields):
...
+        for event in registrations.mapped("event_id"):
+            if event.type != first_type:
+                raise exceptions.ValidationError(
+                    _("You cannot cancel registrations from events of "
+                      "different types at once."))
+        res['event_type_id'] = first_type.id
```
pedrobaeza, line 32: If the first event has no type, this would lead again to error?

event_registration_cancel_reason/wizard/event_registration_cancel_log_reason.py
yajo: Is that even possible?

event_registration_cancel_reason/wizard/event_registration_cancel_log_reason.py
yajo, line 32: No, first_type would be an empty recordset, where all fields are `False`

event_registration_cancel_reason/wizard/event_registration_cancel_log_reason.py
pedrobaeza: Not sure, but making your code bullet-proof is always better

event_registration_cancel_reason/wizard/event_registration_cancel_log_reason.py
pedrobaeza, line 32: OK

event_registration_cancel_reason/wizard/event_registration_cancel_log_reason.py
yajo: Yes, I know, but I feel an exception here would be the right thing to happen. After all, what registrations are you going to cancel if there are none?

pedrobaeza: Put then self.ensure_one()

pedrobaeza: No, but that doesn't serve for multi...

pedrobaeza: I prefer the other formula. It can silently pass to the other window without problems, and it's harmless

yajo: I still cannot find a real use case where you could possibly get to this wizard and `registrations` would be empty (and less where you'd wish the process not to get interrupted). But nevermind, let me change it.

yajo:
Please tag as "needs review".

pedrobaeza:
:+1:

rafaelbn:
Thanks! :+1:

JavierIniesta:
:+1:

PR #42, opened 2016-03-23
[8.0][website_event_filter_selector] Filter events with selectors.

yajo:
You can use selectors above the event list in the website, instead of the left column that there is right now.

@rafaelbn @dgzurita


pedrobaeza:
@Yajo, do you think this can be generalize as a search snippet parameterizing the fields (for any object)? If so, we can gain a lot spending some time on it (although this one gets in this form) for other similar tasks in our projects.

yajo:
I'm pretty sure it's not possible this time. Just read the way the `<option>` tags are built and you'll see its not much standard ORM, because the `/event` controller outputs its fields like that.

Also, controllers seem to search in non-standard ways, I'm not sure if that you ask is even possible.

pedrobaeza:
OK, good to know, but maybe we can make something like injecting values inheriting the controller and calling a method that prepares these values in a standard way, but this is totally out of scope of this PR.

Thanks for the info.

elicoidal:
LGTM :+1:

pedrobaeza:
Mention in "Known issues / Roadmap" that the three filters are activated together, with no possibility of choosing if to hide any of them.

Other than that, :+1:

rafaelbn:
Hi, tested in runbot.

It shoud filter after changing selector and not wait for button search. You can say: "then why do we need search button?" You are right, It's not needed.

It's like this in Odoo defaults filters, you don't have seach button. It just make every search one more click.

Thanks

yajo:
Thanks for your comments.

@pedrobaeza, actually the only one that you cannot avoid is the "date" filter, the others are selectable, just like in core Odoo:

captura de pantalla de 2016-03-29 15-35-23

@rafaelbn Sorry that's the design I got, but you're right, I'm on it.

yajo:
All fixed now.

rafaelbn:
What do you think about this (duplicated filter?)

yajo:
That's probably not what you want, but since both filter styles are opt-in, it could happen. You should just disable the one you don't want and that's it.

rafaelbn:
How to disable?

yajo:
Under "selection filters"

rafaelbn:
But if I disable "Filter by category" it disable both filters. Not only one. I mean to have only one filter by category, in top OR in left column but not both (duplicated)

yajo:
You have "Filters" (core) and "Selection Filters" (this module). Be sure you are (de)selecting the right ones. You can have both or one or none, just choose wisely. :wink:

rafaelbn:
How? I cannot

yajo:
Uncheck "Filters"

rafaelbn:
OK! :+1:

PR #45, opened 2016-04-04
[FIX][8.0][website_event_filter_selector] Do not die without types or countries

### LINKED ISSUE #44: [8][website_event_filter_selector] error when no types found
module: website_event_filter_selector
version: 8.0

**Steps to reproduce**
- remove all event types
- go to events list in website

**Current behavior**
error 500

**Expected behavior**
work

yajo:
This basically fixes #44 and a similar bug detected that would happen when no countries were found.

It just adds a `t-if` to avoid possible failures, [borrowing the idea from core `website_event` module](https://github.com/odoo/odoo/blob/8.0/addons/website_event/views/website_event.xml#L167).

@rafaelbn

### DIFF
--- a/website_event_filter_selector/views/templates.xml
+++ b/website_event_filter_selector/views/templates.xml
@@ -38,15 +38,18 @@
         <label for="country">Where</label>
         <select name="country" class="form-control">
             <t t-foreach="countries" t-as="country">
-                <option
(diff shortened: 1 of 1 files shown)


pedrobaeza:
:+1:

rafaelbn:
Thanks :+1:

PR #47, opened 2016-04-06
[IMP][event_contact] Add contact_ids to event.type.

yajo:
Also it will load the contacts from the event type if the event has none and the type has some.

@rafaelbn @antespi


andhit-r:
Functionality tested on runbot :+1:

pedrobaeza:
:+1:

PR #49, opened 2016-04-27
[8.0][IMP][website_event_filter_selector] Add city filter for events.

yajo:
Since this includes new logic, it adds also a template for left-column filters.

With this patch, you can now filter events depending on their cities, not only on their countries.

This also includes some modifications that will make filters occupy always a full bootstrap row, no matter if there are 1, 2, 3 or 4 filters, to avoid ugly layouts.

@pedrobaeza


rafaelbn:
Tested in runbot :+1:

andhit-r:
Nice. Tested on runbot :+1:

pedrobaeza:
:+1:

PR #50, opened 2016-05-10
[FIX] Add email != False to not_opt_out filter

antespi:
When filtering `event.registration` objects for adding them to a mass.mailing, discard objects with no email defined, because they will fail when sending.

### DIFF
--- a/event_registration_mass_mailing/views/event_registration.xml
+++ b/event_registration_mass_mailing/views/event_registration.xml
@@ -41,8 +41,8 @@
     <field name="arch" type="xml">
         <filter string="New" position="before">
             <filter string="Available for mass mailing"
-                name='not_opt_out' domain="[('opt_out', '=', False)]"
-                help="Registrations that did not ask not to be included in mass mailing campaigns"/>
+                name='not_opt_out' domain="[('opt_out', '=', False), ('email', '!=', False)]"
+                help="Registrations that did not ask not to be included in mass mailing campaigns and has an e-mail set."/>
             <separator/>
         </filter>
     </field>
--- a/event_registration_mass_mailing/__openerp__.py
+++ b/event_registration_mass_mailing/__openerp__.py
@@ -5,7 +5,7 @@
 {
     'name': "Mass mailing from events",
     'category': 'Marketing',
-    'version': '8.0.1.0.0',
+    'version': '8.0.1.1.0',
     'depends': [
         'event',
         'mass_mailing'


rafaelbn:
:+1:

event_registration_mass_mailing/views/event_registration.xml
```diff
@@ -41,7 +41,7 @@
     <field name="arch" type="xml">
         <filter string="New" position="before">
             <filter string="Available for mass mailing"
-                name='not_opt_out' domain="[('opt_out', '=', False)]"
+                name='not_opt_out' domain="[('opt_out', '=', False), ('email', '!=', False)]"
                 help="Registrations that did not ask not to be included in mass mailing campaigns"/>
```
pedrobaeza: Then change this text also for reflecting the correct situation.

antespi: Done, thanks

event_registration_mass_mailing/views/event_registration.xml
```diff
@@ -41,8 +41,8 @@
     <field name="arch" type="xml">
         <filter string="New" position="before">
             <filter string="Available for mass mailing"
-                name='not_opt_out' domain="[('opt_out', '=', False)]"
-                help="Registrations that did not ask not to be included in mass mailing campaigns"/>
+                name='not_opt_out' domain="[('opt_out', '=', False), ('email', '!=', False)]"
+                help="Registrations that did not ask not to be included in mass mailing campaigns and has email defined"/>
```
pedrobaeza: Better to put `and has an e-mail set.`

pedrobaeza:
Please squash the commits and bump version number (the little one).

antespi:
Done and squashed, thanks

pedrobaeza:
Sorry! It's an XML, so it has to be the middle version, because it requires an update.

pedrobaeza:
My fault...

antespi:
Version changed properly

dreispt:
LGTM :+1:

dreispt:
(but it would be nice to have a test case added)

pedrobaeza:
@dreispt, tests don't apply for filters.

Merging.

antespi:
Thanks all for this fast-track merging @dreispt @pedrobaeza

dreispt:
@pedrobaeza yes, you're right.

Issue #51, opened 2016-05-11
[8.0][BUG] website_event_filter_selector - All cities sumatory is incorrect


rafaelbn:
module: website_event_filter_selector
version: 8.0

**Steps to reproduce**
- Go to runbot http://3146931-8-0-3c8cc6.runbot1.odoo-community.org/event
- Make any filter that show you more than 10 events

**Current behavior**
- All Cities filter shows you only 10 cities, it doesnt matter if you have more

**Expected behavior**
- All Cities filter must show the sum of all events

![2016-05-11_2-24-35](https://cloud.githubusercontent.com/assets/8736623/15166524/124824f8-1720-11e6-9ee4-1d65c5544299.jpg)

@yajo @pedrobaeza

PR #52, opened 2016-05-11
[FIX] website_event_filter_selector: Fix date selection

pedrobaeza:
cc @Yajo @antespi @rafaelbn

### DIFF
--- a/website_event_filter_selector/views/templates.xml
+++ b/website_event_filter_selector/views/templates.xml
@@ -22,7 +22,7 @@
             <t t-foreach="dates" t-as="date">
                 <option
                     t-att-value="date[0]"
-                    t-esc="'%s (%s)' % (date[1], date[3])"
+                    t-esc="'%s (%s)' % (date[1], date[3]) if date[0] != 'old' else date[1]"
                     t-att-selected="date[0] == searches.get('date') and
                                     'selected'"/>
             </t>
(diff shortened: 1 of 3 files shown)


rafaelbn:
Fixed :+1: thanks

yajo:
:+1: thanks

website_event_filter_selector/views/templates.xml
```diff
@@ -22,7 +22,7 @@
             <t t-foreach="dates" t-as="date">
                 <option
                     t-att-value="date[0]"
-                    t-esc="'%s (%s)' % (date[1], date[3])"
+                    t-esc="'%s (%s)' % (date[1], date[3]) if date[0] != 'old' else date[1]"
```
antespi, line 25: If date == 'old', then there is only one value for two %s places, isn't it?
Maybe with parenthesis it'll be clearer

```
t-esc="('%s (%s)' % (date[1], date[3])) if date[0] != 'old' else date[1]"
```

pedrobaeza, line 25: No, the expression is not analyzed that way. Operator precedence makes that you don't need parenthesis.

yajo:
This fixes https://github.com/OCA/event/issues/51

website_event_filter_selector/views/templates.xml
antespi, line 25: Well, that's why I point to add parenthesis, to be clearer.
But anyway, this is ok for me, :+1:

PR #55, opened 2016-08-03
[8.0][IMP/FIX][partner_event] More sensible behavior.

yajo:
Before this patch, "Events" button in the partner form displayed the count of its registrations.

After this patch:
- The button displays the count of events, not of registrations.
- A help tooltip is added to help the user understand that.
- Events where all registrations are cancelled or in draft are not taken into account.
- The resulting registrations view defaults to only those confirmed or attended, but the filter can easily be removed.

Also some unused code has been removed (I checked that no OCA addons use it), and tests are updated to reflect that.

If https://github.com/odoo/odoo/pull/12997 gets merged, this will have some code to change, but nothing should break anyway.

@Tecnativa

### DIFF
--- a/partner_event/models/res_partner.py
+++ b/partner_event/models/res_partner.py
@@ -15,7 +15,6 @@ class ResPartner(models.Model):
     event_count = fields.Integer(
         string='Events',
         compute='_compute_event_count',
-        oldname="registration_count",
         help="Count of events with confirmed registrations.",
     )
 
--- a/partner_event/models/res_partner.py
+++ b/partner_event/models/res_partner.py
@@ -3,7 +3,7 @@
 # For copyright and license notices, see __openerp__.py file in root directory
 ##############################################################################
 
-from openerp import models, fields, api
+from openerp import api, fields, models
 
 
 class ResPartner(models.Model):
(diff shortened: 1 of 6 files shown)


partner_event/models/res_partner.py
```diff
@@ -12,19 +12,18 @@ class ResPartner(models.Model):
...
-        compute='_count_attended_registration')
-
-    @api.one
-    @api.depends('registrations')
-    def _count_registration(self):
-        self.registration_count = len(self.registrations)
+    event_count = fields.Integer(
+        string='Events',
+        compute='_compute_event_count',
+        oldname="registration_count",
```
pedrobaeza: As this is not an stored field, you don't need to put oldname

partner_event/models/res_partner.py
```diff
@@ -12,19 +12,18 @@ class ResPartner(models.Model):
...
+        help="Count of events with confirmed registrations.",
+    )
 
     @api.one
     @api.depends('registrations.state')
-    def _count_attended_registration(self):
-        self.attended_registration_count = len(self.registrations.filtered(
-            lambda x: x.state == 'done'))
+    def _compute_event_count(self):
+        self.event_count = len(
```
pedrobaeza, line 24: Have you measured the performance of this?

pedrobaeza:
The drop in coverage is not real (this module is 100% covered).

partner_event/models/res_partner.py
yajo: Doesn't it help in case someone had a module using that old name?

pedrobaeza: No, this only serves to auto-rename the DB column (which doesn't exist)

partner_event/models/res_partner.py
yajo, line 24: I just tested it manually and it seems OK, but not with thousands of records... if you have any suggestion, it's welcome.

yajo:
It must be due to the code reduction.

partner_event/models/res_partner.py
pedrobaeza, line 24: No better option indeed.

pedrobaeza:
:+1:

partner_event/models/res_partner.py
pedrobaeza, line 24: Well, maybe auto_join can help to improve this. Do you know it?

yajo, line 24: It does not apply here AFAIK.

pedrobaeza, line 24: I'm talking about putting that on the link between the registration and the event.

yajo, line 24: I know, but such improvement should be done in Odoo itself if required.

pedrobaeza, line 24: I think it will work if we overwrite the field and put that argument.

yajo, line 24: Yes it will, but if there is a real performance problem there, it will affect not only this module, but all Odoo, and should be patched upstream. That's what I mean.

yajo, line 24: There's still one thing I could do: search_read in `event.registration` and then search by ids directly in the `event.event` model, but still I don't know if that will fix this supposed slowness. Storing the value is another option.

rafaelbn:
Thanks :+1:

PR #56, opened 2016-09-15
[8.0][ADD] website_event_registration_seat_limit module

angelmoya:
Add module that extend event_registration_seat_limit https://github.com/OCA/event/tree/8.0/event_registration_seat_limit to limit the quantity selector on online register process

From: https://github.com/OCA/website/pull/253

### DIFF
--- a/website_event_registration_seat_limit/README.rst
+++ b/website_event_registration_seat_limit/README.rst
@@ -1,0 +1,61 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+============================================
+Seats limits in website events registrations
+============================================
+
+This module was written to extend the functionality of events to support
+setting a maximum and a minimum of seats per registration.
+
+Online registration is limited by maximum and minimum number of seats.
+
+Usage
+=====
+
+To use this module, you need to:
+
+* Enter an event form.
+* Use the new fields in the *Registrations* tab.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/186/8.0
+
+For further information, please visit:
+
+* https://www.odoo.com/forum/help-1
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/event/issues>`_. In
+case of trouble, please check there if your issue has already been reported. If
+you spotted it first, help us smashing it by providing a detailed and welcomed
+feedback `here
+<https://github.com/OCA/event/issues/new?body=module:%20website_event_registration_seat_limit%0Aversion:%208.0.4.0.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+
+
+Credits
+=======
+
+Contributors
+------------
+
+* `Angel Moya <mailto:am@jamotion.ch>`_.
+
+
+Maintainer
+----------
+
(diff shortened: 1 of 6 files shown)


website_event_registration_seat_limit/views/event_template.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<openerp>
+    <data>
+
+        <template id="event_description_full" inherit_id="website_event_sale.event_description_full" name="Event's Ticket limit">
+            <xpath expr="//select[@class='form-control']" position="replace">
+                <select t-if="ticket.seats_available or not ticket.seats_max" t-attf-name="ticket-#{ ticket.id }" class="form-control">
+                    <t t-foreach="range(ticket.event_id.registration_seats_min, min(min(ticket.event_id.registration_seats_max,ticket.seats_available or 9) + 1, 10))" t-as="nb"><option t-esc="nb"/></t>
```
yajo: I think you should check if these constraints apply, because on runbot you cannot chose any quantity.

yajo (changes requested):

website_event_registration_seat_limit/views/event_template.xml
angelmoya: OK, I will check, it's not working with "Maximum of participants per registration" = 0

website_event_registration_seat_limit/README.rst
```diff
@@ -0,0 +1,61 @@
...
+To use this module, you need to:
+
+* Enter an event form.
+* Use the new fields in the *Registrations* tab.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/186/8.0
+
+For further information, please visit:
```
andhit-r, line 25: You don't need this

website_event_registration_seat_limit/__openerp__.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+# © 2016 Jamotion
+
+{
+    "name": "Seats per registration in website events",
+    "version": "8.0.4.0.0",
```
andhit-r, line 7: 8.0.4.0.0? It should be 8.0.1.0.0

andhit-r (approved):
I confirm that with _Maximum of participants per registration_  = 0 user can not chose qt. But it should be fixed on event_registration_seat_limit, not here. So other than minor issue :+1:  from me

yajo:
I'm sorry @andhit-r, but I don't agree. [The field itself explains that you should set it to 0 to ignore the setting](https://github.com/OCA/event/blob/8.0/event_registration_seat_limit/models.py#L16-L17).

andhit-r (changes requested):
@Yajo ups sorry i did not see that. Sorry @angelmoya , you should fix https://github.com/OCA/event/pull/56#discussion_r79576994

angelmoya:
@andhit-r @Yajo It's fixed, I test it on my server, but runbot is not working to test

yajo:
Please remember not to squash every time you patch something. Reviewing a diff is much easier than reviewing again all the code, and mergers have the squash & merge tool for most cases.

pedrobaeza:
The squash should be done at the end of the reviewing process.

yajo:
Bots went :x:

angelmoya:
OK, pep8 fixed, and thanks for the tip

yajo:
Still :x:

rafaelbn:
Hi @angelmoya , thanks for contribution. Please review coverage, could you add tests please? Thanks

PR #57, opened 2016-09-16
[FIX][website_event_excerpt_img] Use event's timezone.

yajo:
Without this patch, the event's start and end dates (and times) will be inaccurate.

@Tecnativa.

### DIFF
--- a/website_event_excerpt_img/__openerp__.py
+++ b/website_event_excerpt_img/__openerp__.py
@@ -4,7 +4,7 @@
 {
     "name": "Excerpt + Image in Events",
     "summary": "New layout for event summary, including an excerpt and image",
-    "version": "8.0.1.0.0",
+    "version": "8.0.1.0.1",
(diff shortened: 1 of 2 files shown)


rafaelbn:
:+1:

website_event_excerpt_img/__openerp__.py
```diff
@@ -4,7 +4,7 @@
 {
     "name": "Excerpt + Image in Events",
     "summary": "New layout for event summary, including an excerpt and image",
-    "version": "8.0.1.0.0",
+    "version": "8.0.1.0.1",
```
pedrobaeza: As this requires an update, 8.0.1.1.0

pedrobaeza (changes requested):

pedrobaeza (approved):

pedrobaeza:
Fast-tracking as needed change.

PR #59, opened 2016-10-07
[NEW][8.0] event_email_reminder: Send reminder events by email

sergio-teruel:
@tecnativa


event_email_reminder/models/event.py
```diff
@@ -0,0 +1,56 @@
...
+from openerp.tools import DEFAULT_SERVER_DATE_FORMAT
+
+
+class EventEvent(models.Model):
+    _inherit = 'event.event'
+
+    @api.model
+    def run_event_email_reminder(
+            self, days=7, near_events=False, template_id=None):
+        """
```
yajo: Usually [1st line of docstring should be the short description of the method](https://www.python.org/dev/peps/pep-0008/#documentation-strings).

event_email_reminder/models/event.py
```diff
@@ -0,0 +1,56 @@
...
+        @param days: number of days to reminder when events star
+        @param near_events: If you want receive the events which start
+        between now and limit date
+        @param template_name: Name of a template or None
+        @return:
```
yajo: I think the standard is to use [Sphinx syntax](http://www.sphinx-doc.org/en/stable/domains.html) for this kind of definitions:

```
:param int days:
    Number of days to reminder when events star
:param bool near_events:
    If you want receive the events which start
etc.
```

event_email_reminder/models/event.py
```diff
@@ -0,0 +1,56 @@
...
+                    'event_email_reminder.event_email_reminder_template')
+            else:
+                template = self.env['email.template'].browse(template_id)
+            for user in events.mapped('user_id'):
+                ctx = self.env.context.copy()
+                event_by_user = events.filtered(lambda x: x.user_id == user)
+                ctx.update({
+                    'events': event_by_user.sorted(lambda x: x.date_begin),
+                })
+                template.with_context(ctx).send_mail(user.partner_id.id)
```
yajo, line 59: I think that instead of all this, you should simply add a new message to the event log, notifying all subscribers.

yajo (changes requested):

yajo:
Did you know an automated action can be just "send an email", and linked to an email template? I think you just need to create automated action + email template + ir.cron and this module should work without any python code.

event_email_reminder/models/event.py
sergio-teruel: The email send a summary of all event that begin on this time, if i use a "message_post" i need do by each event... I do not want this behavior...

yajo: I'm not sure that's actually better. Normally event reminders in all calendar apps are per event, and that way you can notify all subscribers instead of just one (the responsible).
@pedrobaeza your thoughts?

sergio-teruel:
@Yajo Changes done, in next module i will do with automated actions :smirk:

yajo:
Then add to roadmap please.

event_email_reminder/models/event.py
pedrobaeza: One thing: can we use the current Odoo reminders to do the job? Add these reminders to the event record and let act on that way.

yajo: What exactly do you mean by Odoo reminders? I'm not aware of that :confused:

pedrobaeza: The model event.event has a reminder system that you can flexibly define on the frequency, number of alerts, etc

yajo: Are you sure we are talking about events and not about meetings? How do you get to that @pedrobaeza?

pedrobaeza: Sorry, the model is `calendar.event`

sergio-teruel:
@Yajo No in this module. In others.

event_email_reminder/models/event.py
sergio-teruel: @pedrobaeza Are you sure of this??

sergio-teruel: The purpouse of this module is send an email to responsible with a summary of next events.

event_email_reminder/data/event_email_reminder_data.xml
```diff
@@ -0,0 +1,57 @@
...
+
+<record id="ir_cron_event_email_reminder_action" model="ir.cron">
+    <field name="name">Event Email Reminder</field>
+    <field name="active" eval="True"/>
+    <field name="model" eval="'event.event'"/>
+    <field name="function" eval="'run_event_email_reminder'"/>
+    <field name="args" eval="'(5, False, None)'"/>
+    <field name="doall" eval="False"/>
+    <field name="interval_number">1</field>
+    <field name="interval_type">days</field>
```
yajo, line 53: Shouldn't this be `weeks`?

sergio-teruel: why?

yajo: To avoid getting 7 notifications per event, one each day of the week

sergio-teruel: The user can change this if he wants

yajo:
:+1:

event_email_reminder/models/event.py
pedrobaeza: OK, I get it

pedrobaeza (approved):

PR #62, opened 2016-11-07
[ADD] event_kanban_view_optimized

antespi:
Event kanban view optimized
===========================

Event kanban view computes ``is_subscribed`` field per each event and loading time
is degraded when events have many registrations. This addon removes this field
from kanban view in order to optimize loading time.

@Tecnativa


event_kanban_view_optimized/views/event_event_view.xml
```diff
@@ -0,0 +1,20 @@
...
+<data>
+
+<record id="view_event_kanban" model="ir.ui.view">
+    <field name="name">Remove is_subscribed field to improve loading time</field>
+    <field name="model">event.event</field>
+    <field name="inherit_id" ref="event.view_event_kanban"/>
+    <field name="priority" eval="99999"/>
+    <field name="arch" type="xml">
+        <field name="is_subscribed" position="replace"/>
+        <div groups="base.group_user,base.group_portal" position="replace"/>
```
pedrobaeza, line 15: Why this replace?

antespi: This is the section where `is_subscribed` variable is used: https://github.com/odoo/odoo/blob/4809b572835a94543cd1f37e621a98fabec053fd/addons/event/event_view.xml#L227

rafaelbn (approved):
I've tested in production environment of a customers as in runbot there is not such an amount of events an registries.

pedrobaeza (approved):

event_kanban_view_optimized/README.rst
```diff
@@ -0,0 +1,54 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+===========================
+Event kanban view optimized
+===========================
+
+Event kanban view computes ``is_subscribed`` field per each event and loading time
```
yajo: "per event" or "for each event"

yajo (changes requested):

PR #61, opened 2016-11-07
[9.0][event_contact] Migration to v9

mikevhe18:

PR #63, opened 2016-11-08
[9.0][website_event_contact] Migration to V9

mikevhe18:


rafaelbn:
WIP as it depends in https://github.com/OCA/event/pull/61

PR #64, opened 2016-11-12
9.0 mig partner event

cubells:
cc @Tecnativa


pedrobaeza:
One of tests fail (the event_count one).

partner_event/i18n/partner_event.pot
```diff
@@ -1,13 +1,13 @@
 # Translation of Odoo Server.
```
pedrobaeza: Remove this file.

pedrobaeza (changes requested):

pedrobaeza:
Can you include tests for the non tested parts?: https://codecov.io/gh/OCA/event/compare/e07f5051be80c36db2fe0d2d0a87f51a7231e817...1895b898e9d92a8e34d83533f2f8c56108f05fac

partner_event/views/event_registration_view.xml
```diff
@@ -1,31 +1,25 @@
 <?xml version="1.0" encoding="utf-8"?>
-<!-- Copyright 2016 Jairo Llopis <jairo.llopis@tecnativa.com>
```
yajo: Leave this untouched please. Review https://github.com/OCA/maintainer-tools/pull/197#issue-158635253.

partner_event/views/event_registration_view.xml
```diff
@@ -1,31 +1,25 @@
...
-    <field name="name">Search attended registrations</field>
-    <field name="model">event.registration</field>
-    <field name="inherit_id" ref="event.view_registration_search"/>
-    <field name="arch" type="xml">
-        <!-- TODO Remove if https://github.com/odoo/odoo/pull/12997 is merged -->
-        <xpath expr="//filter[@domain=&quot;[('state','=','open')]&quot;]"
-               position="attributes">
-            <attribute name="name">open</attribute>
-        </xpath>
+    <record id="view_registration_search" model="ir.ui.view">
```
yajo: It is common in views to remove 1st indentation level because it adds no value. Those `<odoo>` tags are mandatory anyway, so  they are considered by some as the partially-invisible-but-necessary parts of templates. In fact, [standard templates are like that](https://github.com/OCA/maintainer-tools/blob/master/template/module/views/report_name.xml). These diffs also dificult seeing the real changes.

So, please, here and everywhere, leave indentation as it was if you are changing it just for this reason :wink:

partner_event/__init__.py
```diff
@@ -1,7 +1,5 @@
 # -*- coding: utf-8 -*-
-# © 2014 Serv. Tecnol. Avanzados - Pedro M. Baeza
-# © 2015 Antiun Ingenieria S.L. - Javier Iniesta
-# © 2016 Antiun Ingenieria S.L. - Antonio Espinosa
+# © 2016 Tecnativa S.L. - Vicent Cubells
```
yajo, line 5: Usually it's not fun for people if you remove their copyright and add yours.

You may, however, update its format to last template, or in some of this cases (Pedro & Antonio), change Antiun for Tecnativa, but a good rule of thumb is: do not remove others' copyright, as you probably don't want others to remove yours, unless they ask you to. :wink:

yajo (changes requested):
Some cosmetic/legal changes required.

yajo:
Although my comments are flolded, they are still unfixed.

cubells:
Thanks @Yajo

partner_event/wizard/res_partner_register_event.py
```diff
@@ -26,6 +28,25 @@ def _prepare_registration(self, partner):
...
+        errors = []
         for partner_id in self.env.context.get('active_ids', []):
             partner = partner_obj.browse(partner_id)
-            registration_obj.create(self._prepare_registration(partner))
+            try:
+                with self.env.cr.savepoint():
+                    registration_obj.create(
+                        self._prepare_registration(partner))
+            except:
+                errors.append(partner.name)
```
yajo, line 39: I think it would be nice if you included the error message too, to help user know the reason for it.

Also better if you include a comment on why do you expect that this could fail.

partner_event/views/event_registration_view.xml
pedrobaeza: I prefer to have that indentation level. That's not a guideline, but your preference. Not having this indentation level, IDEs don't work correctly with automatic indentations.

partner_event/wizard/res_partner_register_event.py
```diff
@@ -26,6 +28,25 @@ def _prepare_registration(self, partner):
...
+                                    'res_partner_register_event_view')
+            return {
+                'type': 'ir.actions.act_window',
+                'res_model': self._name,
+                'view_mode': 'form',
+                'view_type': 'form',
+                'view_id': [data_obj.id],
+                'res_id': self.id,
+                'target': 'new',
+            }
```
yajo, line 52: Is it really necessary to open this same form when you click on the button? Is it not open already?

partner_event/wizard/res_partner_register_event_view.xml
```diff
@@ -1,25 +1,38 @@
...
-                <button special="cancel"
-                        string="Cancel"
-                        />
+                <div attrs="{'invisible': [('errors', '!=', False)]}">
+                    <button name="button_register"
+                            type="object"
+                            string="Create registrations"
+                            class="oe_highlight"
+                            />
+                    or
```
yajo: I think these "or"s are being removed everywhere. Less unnecessary translation strings.

partner_event/wizard/res_partner_register_event_view.xml
```diff
@@ -1,25 +1,38 @@
...
+                    <button special="cancel"
+                            string="Cancel"
+                            />
+                </div>
+                <div attrs="{'invisible': [('errors', '=', False)]}">
+                    <button special="cancel"
+                            string="OK"
```
yajo: I think you can drop this. Users already have the X in the dialog, so not really adding anything interesting IMHO.

partner_event/models/event_registration.py
```diff
@@ -0,0 +1,52 @@
...
+                # Create partner
+                partner = partner_model.sudo().create(
+                    self._prepare_partner(vals))
+                partner_id = partner.id
+            vals['partner_id'] = partner_id
+        return super(EventRegistration, self).create(vals)
+
+    @api.multi
+    def partner_data_update(self, data):
+        reg_fields = ['name', 'email', 'phone']
```
yajo: Normally if you need to use a list only for checking existence with the `in` operator, it is better to use sets. Not very critical in a list of 3, so not blocking, but a good point to have in mind in general :wink:.

yajo (dismissed):
OK, now digging into code, a little requests.

partner_event/views/event_registration_view.xml
yajo: But certainly diffs are easier to interpret. I usually try not to change it too when somebody wrote it in a different style than mine :grin:

partner_event/wizard/res_partner_register_event.py
pedrobaeza: Yes, it is. Odoo if not closes the dialog

partner_event/wizard/res_partner_register_event_view.xml
pedrobaeza: It's an Odoo guideline to include. More user friendly and more easily clickable.

partner_event/wizard/res_partner_register_event.py
yajo: Have you tried returning `True`? Not sure where, I saw that happens when you return `False`, that might be the default...

pedrobaeza: It's closed independently of what you return if I don't remember bad, but we can try.

pedrobaeza (approved):

PR #74, opened 2017-04-05
[FIX][website_event_filter_selector] Fix Travis build

yajo:
Builds failed on 2017-03-31 because the tour was expecting to see USA event starting in this month, but next day was another month.

Well, in short, this fix makes the tour work on any day.

@Tecnativa


pedrobaeza:
Travis is still failing.

pedrobaeza (approved):
Fast-tracking to fix branch

PR #78, opened 2017-06-02
[10.0][NEW] event_registration_multi_qty: New module for allow registration writing quantities.

sergio-teruel:
cc @Tecnativa

### DIFF
--- a/event_registration_multi_qty/README.rst
+++ b/event_registration_multi_qty/README.rst
@@ -8,29 +8,32 @@ Event Registration Multi Qty
 
 This module allows to make registration to events with more than one attendee.
 
+This was the unique behavior on versions 8 and 9. As it's useful for some use
+cases, this module allows both ways of registering attendees.
+
 
 Configuration
 =============
 
-.. figure:: path/to/local/image.png
-   :alt: alternative description
-   :width: 600 px
+To configure the possibility of registering mutiple attendees:
+
+#. Go to *Events > Events*.
+#. Create one event.
+#. Go to *Registration tab* and check "Registration Multi Qty" to activate
+   this function.
 
 Usage
 =====
 
-To use this module, you need to:
+For registering multiple attendees:
 
-#. Go to *Events* and create one.
-#. Go to *Registration tab* and check "Registration Multi Qty" to activate
-   this function.
-#. Create an attendee for this event and you can write the quantities for this
-   this registration.
+#. Create an attendee for the created event.
+#. Write the quantities for this registration.
 
 
 .. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
    :alt: Try me on Runbot
-   :target: https://runbot.odoo-community.org/runbot/199/10
+   :target: https://runbot.odoo-community.org/runbot/199/10.0
 
 
 Bug Tracker
@@ -53,6 +56,7 @@ Contributors
 ------------
 
 * Sergio Teruel <sergio.teruel@tecnativa.com>
+* Pedro M. Baeza <pedro.baeza@tecnativa.com>
 
 
 Maintainer
--- a/event_registration_multi_qty/README.rst
+++ b/event_registration_multi_qty/README.rst
@@ -1,0 +1,71 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+============================
+Event Registration Multi Qty
+============================
+
+This module allows to make registration to events with more than one attendee.
+
+
+Configuration
+=============
+
+.. figure:: path/to/local/image.png
+   :alt: alternative description
+   :width: 600 px
+
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to *Events* and create one.
+#. Go to *Registration tab* and check "Registration Multi Qty" to activate
+   this function.
+#. Create an attendee for this event and you can write the quantities for this
+   this registration.
+
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/199/10
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/event/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smash it by providing detailed and welcomed feedback.
+
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Sergio Teruel <sergio.teruel@tecnativa.com>
+
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
--- a/event_registration_multi_qty/__manifest__.py
+++ b/event_registration_multi_qty/__manifest__.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Sergio Teruel<sergio.teruel@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Event Registration Multi Qty',
+    'version': '10.0.1.0.0',
+    'author': 'Tecnativa, '
+              'Odoo Community Association (OCA)',
+    "license": "AGPL-3",
+    'website': 'https://www.tecnativa.com',
+    'category': 'Marketing',
+    'summary': 'Allow registration grouped by quantities',
+    'depends': ['event'],
+    'data': [
+        'views/event_view.xml',
+    ],
+    'installable': True,
+}
--- a/event_registration_multi_qty/models/event.py
+++ b/event_registration_multi_qty/models/event.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
-# Copyright 2017 Sergio Teruel<sergio.teruel@tecnativa.com>
+# Copyright 2017 Tecnativa - Sergio Teruel
+# Copyright 2017 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import _, api, fields, models
@@ -18,9 +19,8 @@ class EventEvent(models.Model):
     @api.depends('seats_max', 'registration_ids.state',
                  'registration_ids.qty')
     def _compute_seats(self):
-        for event in self:
-            if not event.registration_multi_qty:
-                return super(EventEvent, self)._compute_seats()
+        multi_qty_events = self.filtered('registration_multi_qty')
+        for event in multi_qty_events:
             vals = {
                 'seats_unconfirmed': 0,
                 'seats_reserved': 0,
@@ -45,6 +45,8 @@ class EventEvent(models.Model):
                 vals['seats_unconfirmed'] + vals['seats_reserved'] +
                 vals['seats_used'])
             event.update(vals)
+        rest = self - multi_qty_events
+        super(EventEvent, rest)._compute_seats()
 
     @api.multi
     @api.constrains('registration_multi_qty')
@@ -54,7 +56,7 @@ class EventEvent(models.Model):
                     max(event.registration_ids.mapped('qty') or [0]) > 1:
                 raise ValidationError(
                     _('You can not disable this option if there are '
-                      'registrations with quantities grather than one.'))
+                      'registrations with quantities greater than one.'))
 
 
 class EventRegistration(models.Model):
--- a/event_registration_multi_qty/models/event.py
+++ b/event_registration_multi_qty/models/event.py
@@ -1,0 +1,78 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Sergio Teruel<sergio.teruel@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class EventEvent(models.Model):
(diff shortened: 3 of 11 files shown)


event_registration_multi_qty/README.rst
```diff
@@ -0,0 +1,56 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+============================
+Event Registration Multi Qty
```
yajo: IMHO `event_registration_qty` should be enough. The same for this title.

event_registration_multi_qty/README.rst
```diff
@@ -0,0 +1,56 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
```
yajo: Could you update to [latest template](https://raw.githubusercontent.com/OCA/maintainer-tools/master/template/module/README.rst) please?

event_registration_multi_qty/README.rst
```diff
@@ -0,0 +1,56 @@
...
+
+============================
+Event Registration Multi Qty
+============================
+
+This module allows to make registration to events with more than one quantity.
```
yajo: "more than one quantity" -> "more than one attendee".

event_registration_multi_qty/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+# Copyright 2017 Sergio Teruel<sergio.teruel@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Event Registration Multi Qty',
+    'version': '10.0.1.0.0',
+    'author': 'Tecnativa, '
+              'Odoo Community Association (OCA)',
+    "license": "AGPL-3",
+    'website': 'https://odoo-community.org/',
```
yajo: https://www.tecnativa.com

event_registration_multi_qty/models/event.py
```diff
@@ -0,0 +1,58 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class EventEvent(models.Model):
+    _inherit = 'event.event'
+
+    registration_multi_qty = fields.Boolean(
+        string='Registration Multi Qty',
```
yajo: "Allow multiple attendees per registration"

event_registration_multi_qty/models/event.py
```diff
@@ -0,0 +1,58 @@
...
+        help='Allow more than one registration'
```
yajo: With the above `string` you do not need this `help` I think

event_registration_multi_qty/models/event.py
```diff
@@ -0,0 +1,58 @@
...
+    )
+
+    @api.onchange('registration_multi_qty')
+    def _onchange_registration_multi_qty(self):
+        if not self.registration_multi_qty:
+            self.registration_ids.write({'qty': 1})
```
yajo: Calling `write` under an `onchange` is dangerous, because onchanges work in a draft record while `write` don't. You should call `update` instead.

Anyway, this does not seem to be a good choice. If somebody clicks here by mistake, he'd be overriding all qtys. I feel for this is better to add a `constraint`, so that if you already have registrations with `qty != 1` you cannot check this box. The same constraint could be called when changing `qty` field in registrations.

----

Further reading showed me that [after all, `qty` is ignored if disabled](https://github.com/OCA/event/pull/78/files#diff-e670b96f19d2dfc8404e228ec4a39b0aR27), so why even bother about this?

event_registration_multi_qty/models/event.py
```diff
@@ -0,0 +1,58 @@
...
+
+    @api.multi
+    @api.depends('seats_max', 'registration_ids.state',
+                 'registration_ids.qty')
+    def _compute_seats(self):
+        for event in self:
+            if not event.registration_multi_qty:
+                return super(EventEvent, self)._compute_seats()
+            event.seats_unconfirmed = event.seats_reserved = 0
+            event.seats_used = event.seats_available = 0
```
yajo: In these 2 lines you are actually performing 4 writes to DB. Later below, each `+=` call performs another write. How about working with simple variables that start with `0` and perform a single `.write({...})` at the end? 🤔

event_registration_multi_qty/reports/report_event_registration.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class ReportEventRegistration(models.Model):
+    """Events Analysis"""
+    _inherit = "report.event.registration"
+
+    def _sub_select(self):
+        select_str = super(ReportEventRegistration, self)._sub_select()
+        select_str = select_str.replace('count(r.event_id)', 'sum(qty)')
+        select_str = select_str.replace('count(r.id)', 'sum(qty)')
```
Yajo: If you decide to remove the constraint & onchange as said above, then you'd have to keep it in mind here.

event_registration_multi_qty/models/event.py
```diff
@@ -0,0 +1,58 @@
...
+class EventRegistration(models.Model):
+    _inherit = 'event.registration'
+
+    qty = fields.Integer(
+        string='Quantity',
+        required=True,
+        default=1,
+    )
+    registration_multi_qty = fields.Boolean(
+        related='event_id.registration_multi_qty')
```
yajo: I'd add `readonly=True` here too.

yajo (approved):
I have a lot of suggested improvements, but none of them seem to be blocking. Code review OK.

PS: I guess this will have to be combined with a `website_event_sale_registration_qty` addon, right?

sergio-teruel:
@Yajo Changes done, no changes in module name

event_registration_multi_qty/models/event.py
```diff
@@ -0,0 +1,73 @@
...
+                elif registration['state'] == 'done':
+                    event.seats_used += registration['qty']
+            if event.seats_max > 0:
+                event.seats_available = event.seats_max - (
+                    event.seats_reserved + event.seats_used)
+            event.seats_expected = (
+                event.seats_unconfirmed + event.seats_reserved +
+                event.seats_used)
+
+    @api.multi
```
yajo: AFAIK this is no more needed in v10. Not a requirement though.

event_registration_multi_qty/models/event.py
```diff
@@ -0,0 +1,73 @@
...
+    @api.constrains('registration_multi_qty')
+    def _check_attendees_qty(self):
+        for event in self:
+            if not event.registration_multi_qty and \
+                    max(event.registration_ids.mapped('qty') or [0]) > 1:
+                raise ValidationError(
+                    _('You can not inactive this option if there is '
```
yajo: inactive > disable
is > are

event_registration_multi_qty/models/event.py
```diff
@@ -0,0 +1,73 @@
...
+
+    @api.multi
+    @api.constrains('qty')
+    def _check_attendees_qty(self):
+        for registration in self:
+            if not registration.event_id.registration_multi_qty and \
+                    max(registration.event_id.registration_ids.mapped(
+                        'qty') or [0]) > 1:
+                raise ValidationError(
+                    _('You can not add quantities if you not active the'
```
yajo: active > activate

event_registration_multi_qty/models/event.py
```diff
@@ -0,0 +1,73 @@
...
+        default=1,
+    )
+
+    @api.multi
+    @api.constrains('qty')
+    def _check_attendees_qty(self):
+        for registration in self:
+            if not registration.event_id.registration_multi_qty and \
+                    max(registration.event_id.registration_ids.mapped(
+                        'qty') or [0]) > 1:
```
yajo: This check is IMHO overloaded. While editing a registration, you don't need to check all others from the same event, since they don't changed.

`if True in self.mapped("event_id.registration_multi_qty") and self.filtered(lambda reg: reg.qty != 1)` should be enough

event_registration_multi_qty/reports/report_event_registration.py
Yajo: You added the constraint so this is 👍 now 😉

yajo (changes requested):

yajo:
Check Travis too please

sergio-teruel:
@Yajo Changes done

yajo (approved):

event_registration_multi_qty/views/event_view.xml
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0"?>
+<odoo>
+
+<record id="view_event_form" model="ir.ui.view">
+    <field name="model">event.event</field>
+    <field name="inherit_id" ref="event.view_event_form"/>
+    <field name="arch" type="xml">
+        <field name="auto_confirm" position="after">
+            <field name="registration_multi_qty"/>
```
chienandalu: Not a good hook since event_sale overrides it. Any better position?

sergio-teruel:
@chienandalu changes done

event_registration_multi_qty/README.rst
```diff
@@ -0,0 +1,71 @@
...
+.. figure:: path/to/local/image.png
+   :alt: alternative description
+   :width: 600 px
+
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to *Events* and create one.
```
pedrobaeza: These are configuration steps, not usage steps.

event_registration_multi_qty/README.rst
```diff
@@ -0,0 +1,71 @@
...
+#. Go to *Registration tab* and check "Registration Multi Qty" to activate
+   this function.
+#. Create an attendee for this event and you can write the quantities for this
```
pedrobaeza: This one should be kept here though.

event_registration_multi_qty/models/event.py
```diff
@@ -0,0 +1,78 @@
...
+    registration_multi_qty = fields.Boolean(
+        string='Allow multiple attendees per registration',
+        default=True,
+    )
+
+    @api.multi
+    @api.depends('seats_max', 'registration_ids.state',
+                 'registration_ids.qty')
+    def _compute_seats(self):
+        for event in self:
```
pedrobaeza: Better for performance reasons on super: 
```
multi_qty_events = self.filtered('registration_multi_qty')
for event in multi_qty_events:
    ...
rest = self - multi_qty_events
super(EventEvent, rest)._compute_seats()
```

event_registration_multi_qty/models/event.py
```diff
@@ -0,0 +1,78 @@
...
+            if not event.registration_multi_qty:
+                return super(EventEvent, self)._compute_seats()
```
pedrobaeza: Take care with this next time! (although now with the new code proposed this dissapears). This would end the computation in the middle if you find an event without multi qty.

event_registration_multi_qty/models/event.py
```diff
@@ -0,0 +1,78 @@
...
+
+    @api.multi
+    @api.constrains('registration_multi_qty')
+    def _check_attendees_qty(self):
+        for event in self:
+            if not event.registration_multi_qty and \
+                    max(event.registration_ids.mapped('qty') or [0]) > 1:
+                raise ValidationError(
+                    _('You can not disable this option if there are '
+                      'registrations with quantities grather than one.'))
```
pedrobaeza: s/grather/greater (and change source translation also).

event_registration_multi_qty/models/event.py
```diff
@@ -0,0 +1,78 @@
...
+                    vals['seats_reserved'] + vals['seats_used'])
+            vals['seats_expected'] = (
+                vals['seats_unconfirmed'] + vals['seats_reserved'] +
+                vals['seats_used'])
+            event.update(vals)
+
+    @api.multi
+    @api.constrains('registration_multi_qty')
+    def _check_attendees_qty(self):
+        for event in self:
```
pedrobaeza, line 54: `for event in self.filtered(lambda x: not x.registration_multi_qty):` and remove the condition inside the loop.

event_registration_multi_qty/models/event.py
```diff
@@ -0,0 +1,78 @@
...
+    qty = fields.Integer(
+        string='Quantity',
+        required=True,
+        default=1,
+    )
+
+    @api.multi
+    @api.constrains('qty')
+    def _check_attendees_qty(self):
+        for registration in self:
```
pedrobaeza, line 74: Put directly:
`if self.filtered(lambda x: not x.event_id.registration_multi_qty and registration.qty > 1):`

event_registration_multi_qty/reports/report_event_registration.py
```diff
@@ -0,0 +1,16 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import models
+
+
+class ReportEventRegistration(models.Model):
+    """Events Analysis"""
+    _inherit = "report.event.registration"
+
+    def _sub_select(self):
```
pedrobaeza: This won't work if https://github.com/odoo/odoo/pull/16633 is not merged, so we have to wait.

pedrobaeza (changes requested):

event_registration_multi_qty/README.rst
yajo: I think these are usage steps: you need to create an event to use this addon.

event_registration_multi_qty/README.rst
```diff
@@ -0,0 +1,71 @@
...
+   this registration.
+
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/199/10
```
yajo: Broken link.

event_registration_multi_qty/models/event.py
Yajo: Other option: `for event in self - self.filtered("registration_multi_qty"):`

event_registration_multi_qty/models/event.py
Yajo: Typo: `x.qty` instead. But I agree. 😉

yajo (changes requested):

event_registration_multi_qty/models/event.py
yajo: Oh well, thiking it twice... 🤔  @sergio-teruel is being more performant here, because the loop stops on the first match, so he better leaves this as it is. 👍

pedrobaeza: OK

pedrobaeza (approved):

pedrobaeza:
@Yajo, please give your final blessing.

yajo (approved):
Awesome 👍

chienandalu (approved):
Reports don't count properly on runbot (new report query not yet propagated?). Tested locally with success :smiley:

PR #84, opened 2017-06-23
[MIG] website_event_type_description: Migration to 10.0

chienandalu:
Event Type Description in Website
=================================

This module extends the functionality of events website to support a custom
description for each event type that will be displayed only when events of that
type are filtered, below the common description for the Events page.

cc @Tecnativa


chienandalu:
@Yajo How would you better implement this on 10.0 website? Right now it isn't obvious that there is an editable area for the event types...

pedrobaeza (approved):
Indeed it's not obvious from the user perspective as there are no visual hints, but it works (tried in runbot)! Add a comment in "Known issues" about that, and we can go on.

chienandalu:
Ok. Readme edited and commits squashed.

yajo:
I'd rather go for putting a `<div class="oe_editable oe_empty" t-field="current_category.description"/>` and let users edit it directly through the website editor. Is it possible?

pedrobaeza:
I was able to add the block in the website editor in runbot...

yajo (approved):
Oops, true; my fault.

chienandalu:
Yes, I was wondering if there is like a sort of placeholder for this cases whenever editing is active.

yajo:
If you add `class="oe_editable oe_empty"` to the div, I think it adds the default placeholder for you.

chienandalu:
I tried with that but the behavior was the same...

yajo:
Then it seems they removed the feature.

PR #85, opened 2017-06-23
[10.0][MIG][website_event_filter_selector]

yajo:
Includes update to use new tours api.

@Tecnativa


pedrobaeza:
Testing on runbot, if I filter by city, the rest of the combos is not updated:

seleccion_003

That only happens with this combo.

yajo:
It should be fixed now, good catch!

chienandalu (approved):
Functional review :+1:

luismontalba (approved):
Tested on runbot

yajo:
@pedrobaeza merge?

pedrobaeza:
Squash first

yajo:
I squashed old commits a bit also, all of them were mine after all. Migration is now only in 1.

PR #88, opened 2017-07-03
[IMP] Fix pending TODO and add known issue

chienandalu:
Known Issues / Roadmap
----------------------

* In registration contact field has no onchange check so changing the partner
  is not going to change any info on the registration.

cc @Tecnativa


pedrobaeza (changes requested):
Explain also in the README the update from partner to registration on future events

chienandalu:
@pedrobaeza Amended

pedrobaeza (approved):

luismontalba (approved):
Tested on runbot

PR #94, opened 2017-10-23
[IMP] partner_event: forbid partner deletion if it has event registrations

chienandalu:
- If the partner has event registrations it won't be allowed to delete it.

cc @Tecnativa

### DIFF
--- a/partner_event/README.rst
+++ b/partner_event/README.rst
@@ -16,6 +16,7 @@ It also includes:
 * Partner column is visible on registration one2many list inside the event.
 * Action in partner tree view 'More' button, to register several partners
   to an event
+* Restricts partner deletion when event attendees are linked to it.
 
 Configuration
 =============
--- a/partner_event/models/event_registration.py
+++ b/partner_event/models/event_registration.py
@@ -11,6 +11,9 @@ from odoo import api, fields, models
 class EventRegistration(models.Model):
     _inherit = "event.registration"
 
+    partner_id = fields.Many2one(
+        ondelete='restrict',
+    )
     attendee_partner_id = fields.Many2one(
         comodel_name='res.partner',
         string='Attendee Partner',
--- a/partner_event/tests/test_event_registration.py
+++ b/partner_event/tests/test_event_registration.py
@@ -3,17 +3,19 @@
 # © 2015 Tecnativa S.L. - Javier Iniesta
 # © 2016 Tecnativa S.L. - Antonio Espinosa
 # © 2016 Tecnativa S.L. - Vicent Cubells
+# Copyright 2017 Tecnativa - David Vidal
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
+from psycopg2 import IntegrityError
 from datetime import datetime, timedelta
 from odoo.tests import common
 from odoo import fields
 
 
 class TestEventRegistration(common.SavepointCase):
-
-    def setUp(cls):
-        super(TestEventRegistration, cls).setUp()
+    @classmethod
+    def setUpClass(cls):
+        super(TestEventRegistration, cls).setUpClass()
         cls.event_0 = cls.env['event.event'].create({
             'name': 'Test event',
             'date_begin': fields.Datetime.now(),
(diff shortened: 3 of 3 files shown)


partner_event/tests/test_event_registration.py
```diff
@@ -73,3 +75,15 @@ def test_data_update(self):
...
+    def test_unlink(self):
+        # We can't delete a partner with registrations
+        with self.assertRaises(UserError):
+            self.partner_01.unlink()
+        # Create a brand new partner and delete it
+        partner3 = self.env['res.partner'].create({
+            'name': 'unregistered partner'
+        })
+        partner3.unlink()
+        with self.assertRaises(MissingError):
```
pedrobaeza: Do `self.assertFalse(partner3.exists())` instead

partner_event/tests/test_event_registration.py
```diff
@@ -73,3 +75,15 @@ def test_data_update(self):
...
         self.assertEqual(
             event_2.registration_ids.email, 'new@test.com')
+
+    def test_unlink(self):
+        # We can't delete a partner with registrations
+        with self.assertRaises(UserError):
+            self.partner_01.unlink()
+        # Create a brand new partner and delete it
+        partner3 = self.env['res.partner'].create({
+            'name': 'unregistered partner'
```
pedrobaeza: Always comma at the end for not having a 2 lines diff in the future

partner_event/models/res_partner.py
```diff
@@ -54,3 +55,11 @@ def write(self, data):
         res = super(ResPartner, self).write(data)
         self.mapped('registrations').partner_data_update(data)
         return res
+
+    @api.multi
+    def unlink(self):
+        if self.registration_count > 0:
```
pedrobaeza: No need of `> 0`

pedrobaeza: But I think you can just put on the one2many ondelete='restrict'. I'm not sure if it's going to work, but you have tests for that.

pedrobaeza (changes requested):
Change module version

chienandalu:
@pedrobaeza Changes done

chienandalu:
@rafaelbn Please review

pedrobaeza (approved):

pedrobaeza:
Indeed the best option is to restrict on the many2one instead.

rafaelbn:
@chienandalu This Improvement was already done in v8 in : https://github.com/OCA/event/commit/9dc9d383856b7dff5b9db938797e4cb9e1a2a91c

But that changes were lost in v9 migration and in v10 migration.

This means that we loose also the code refactoring that @pedrobaeza did

I think we should recover what it was done

chienandalu:
@rafaelbn Changes done

partner_event/README.rst
```diff
@@ -16,6 +16,7 @@ It also includes:
 * Partner column is visible on registration one2many list inside the event.
```
rafaelbn, line 16: This is wrong @chienandalu as this list doesn't exist any more

partner_event/models/event_registration.py
```diff
@@ -30,15 +34,14 @@ def create(self, vals):
                 [('email', '=ilike', email)], limit=1)
             event = event_model.browse(vals['event_id'])
             if partner:
-                partner_id = partner.id
                 vals['name'] = vals.get('name') or partner.name
                 vals['phone'] = vals.get('phone') or partner.phone
             elif event.create_partner:
                 # Create partner
```
rafaelbn: @chienandalu hello! This doesn't work if you register in a free event. The partner is not created. Should we fix it? I've seen two different behavior for the same workflow (registering a partner) depending if the register is executed in the backend or in the website

rafaelbn (approved):
I've tested "partner_event: forbid partner deletion if it has event registrations" and it works fine 👍 But I have found some inconsistencies in this module. We can merge this IMP and make a FIX for the other staff.

partner_event/__manifest__.py
```diff
@@ -8,7 +8,7 @@
 
 {
     'name': 'Link partner to events',
-    'version': '10.0.1.0.0',
+    'version': '10.0.1.0.1',
```
yajo: I'm not sure, but I think this should be `10.0.1.1.0`, since this is a new feature.

partner_event/tests/test_event_registration.py
```diff
@@ -73,3 +75,14 @@ def test_data_update(self):
         self.partner_01.write({'email': 'new@test.com'})
         self.assertEqual(
             event_2.registration_ids.email, 'new@test.com')
+
+    def test_delete_registered_partner(self):
+        # We can't delete a partner with registrations
+        with self.assertRaises(IntegrityError), self.cr.savepoint():
+            self.partner_01.unlink()
+        # Create a brand new partner and delete it
```
yajo, line 94: I think it makes more sense to delete `partner_01` registrations and then try to delete it again, for the test purposes.

yajo (approved):
Some non-blockings, code OK

rafaelbn:
Please @chienandalu rebase to check of all goes green, thanks!

PR #95, opened 2017-10-31
[10.0][NEW] website_event_questions_template: Templates for your event questions

sergio-teruel:
cc @Tecnativa


website_event_questions_template/README.rst
```diff
@@ -0,0 +1,60 @@
...
+
+This module extends website_event_questions and allow to define questions
+templates for your events.
+
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to *Events > Configuration > Questions Templates* and create all
```
pedrobaeza: This should go in Configuration section, not Usage one.

website_event_questions_template/README.rst
```diff
@@ -0,0 +1,60 @@
...
+check there if your issue has already been reported. If you spotted it first,
+help us smash it by providing detailed and welcomed feedback.
+
+Credits
+=======
+
+Contributors
+------------
+
+* Sergio Teruel <sergio.teruel@tecnativa.com>
```
pedrobaeza: Put a new format as now company URL is not in the manifest:
```rst
* Tecnativa (https://www.tecnativa.com)
  * Sergio Teruel <sergio.teruel@tecnativa.com>
```

website_event_questions_template/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+{
+    'name': 'Website Event Questions Template',
+    'summary': 'Set question templates for events',
+    'version': '10.0.1.0.0',
+    'author': 'Tecnativa, '
+              'Odoo Community Association (OCA)',
+    "license": "LGPL-3",
+    'website': 'https://odoo-community.org/',
```
pedrobaeza: https://github.com/OCA/event

website_event_questions_template/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Sergio Teruel <sergio.teruel@tecnativa.com>
```
pedrobaeza: Better to remove here the e-mail, as in the header, and you have only one place where updating it:
```python
# Copyright 2017 Tecnativa - Sergio Teruel
```

website_event_questions_template/i18n/es.po
```diff
@@ -0,0 +1,185 @@
...
+#: model:ir.model.fields,field_description:website_event_questions_template.field_event_question_template_answer_id
+#: model:ir.model.fields,field_description:website_event_questions_template.field_event_question_template_id
+#: model:ir.model.fields,field_description:website_event_questions_template.field_event_question_template_question_id
+msgid "ID"
+msgstr "ID (identificación)"
+
+#. module: website_event_questions_template
+#: model:ir.model.fields,help:website_event_questions_template.field_event_question_template_question_is_individual
+msgid ""
+"If True, this question will be asked for every attendee of a reservation. If "
```
pedrobaeza: You still continue using POEdit with the line wrapping option, which totally breaks the PO file. Please don't use it or remove that option.

website_event_questions_template/i18n/es.po
```diff
@@ -0,0 +1,185 @@
...
+"not it will be asked only once and its value propagated to every attendees."
+msgstr ""
+"Si esta marcada, esta pregunta se le pedirá a cada participante de una "
+"\"\"reserva. Si no, se le preguntará sólo una vez y su valor se propagará a "
```
pedrobaeza: This is also a side effect of using POEdit.

website_event_questions_template/i18n/es.po
```diff
@@ -0,0 +1,185 @@
...
+msgid "ID"
+msgstr "ID (identificación)"
+
+#. module: website_event_questions_template
+#: model:ir.model.fields,help:website_event_questions_template.field_event_question_template_question_is_individual
+msgid ""
+"If True, this question will be asked for every attendee of a reservation. If "
+"not it will be asked only once and its value propagated to every attendees."
+msgstr ""
+"Si esta marcada, esta pregunta se le pedirá a cada participante de una "
```
pedrobaeza: s/esta marcada/está marcada

website_event_questions_template/i18n/es.po
```diff
@@ -0,0 +1,185 @@
...
+#. module: website_event_questions_template
+#: model:ir.model.fields,field_description:website_event_questions_template.field_event_question_template_name
+msgid "Name"
+msgstr "Nombre"
+
+#. module: website_event_questions_template
+#: model:ir.model.fields,field_description:website_event_questions_template.field_event_question_template_answer_question_id
+#, fuzzy
+msgid "Question id"
+msgstr "Id cuenta"
```
pedrobaeza: This is because you haven't put any string on this reference. Better to put one, but at least, translate it as `Id de pregunta`

website_event_questions_template/i18n/es.po
```diff
@@ -0,0 +1,185 @@
...
+#. module: website_event_questions_template
+#: model:ir.model.fields,field_description:website_event_questions_template.field_event_question_template_question_title
+msgid "Title"
+msgstr "Título"
+
+#. module: website_event_questions_template
+#: code:addons/website_event_questions_template/models/event.py:29
+#, python-format
+msgid "You can not load a template if there are defined questions"
+msgstr "No puedes cargar una plantilla el evento ya tiene preguntas definidas"
```
pedrobaeza: s/plantilla el evento/plantilla **si** el evento

website_event_questions_template/i18n/es.po
```diff
@@ -0,0 +1,185 @@
...
+msgid "event.question.template.answer"
+msgstr "Responder pregunta relativa al puesto"
+
+#. module: website_event_questions_template
+#: model:ir.model,name:website_event_questions_template.model_event_question_template_question
+#, fuzzy
+msgid "event.question.template.question"
+msgstr "Plantilla de preguntas de eventos"
+
+#~ msgid "Question Template"
```
pedrobaeza: Why is this commented? Other side effect of POEdit?

website_event_questions_template/i18n/es.po
```diff
@@ -0,0 +1,185 @@
...
+
+#. module: website_event_questions_template
+#: model:ir.model,name:website_event_questions_template.model_event_config_settings
+msgid "event.config.settings"
+msgstr "Abrir menú 'Configuración'"
+
+#. module: website_event_questions_template
+#: model:ir.model,name:website_event_questions_template.model_event_question_template
+#, fuzzy
+msgid "event.question.template"
```
pedrobaeza: Why don't you put a description in the model also instead of only put the translation?

website_event_questions_template/models/__init__.py
```diff
@@ -0,0 +1,6 @@
+# -*- coding: utf-8 -*-
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from . import res_config
```
pedrobaeza: res_config is a TransientModel. It should go in `wizards` folder.

website_event_questions_template/models/event.py
```diff
@@ -0,0 +1,42 @@
...
+
+    event_question_template_id = fields.Many2one(
+        comodel_name='event.question.template',
+        string='Questions Template',
+        default=lambda d: d._default_question_template_id(),
+    )
+
+    def load_question_template(self):
+        # Onchange method is not possible due to
+        # https://github.com/odoo/odoo/issues/20163#issuecomment-340593894
```
pedrobaeza: Remove last part of the URL (from `#` to the end)

website_event_questions_template/models/event_question_template.py
```diff
@@ -0,0 +1,59 @@
...
+# Copyright 2017 Sergio Teruel <sergio.teruel@tecnativa.com>
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import fields, models
+
+
+class EventQuestionTemplateQuestion(models.Model):
+    _name = 'event.question.template.question'
+
+    title = fields.Char(required=True, translate=True)
```
pedrobaeza: Please inherit from `event.question` changing `_name` and the one2many field:
```python
class EventQuestionTemplateQuestion(models.Model):
    _inherit = 'event.question'
    _name = 'event.question.template.question'

    answer_ids = fields.One2many(
        comodel_name='event.question.template.answer',
    )
```

website_event_questions_template/models/event_question_template.py
```diff
@@ -0,0 +1,59 @@
...
+        required=True,
+        ondelete='cascade',
+    )
+
+
+class EventQuestionTemplateAnswer(models.Model):
+    _name = 'event.question.template.answer'
+    _order = 'sequence,id'
+
+    name = fields.Char(
```
pedrobaeza: Inherit the same way from `event.answer`. With this, new fields added via extra modules are also added here.

website_event_questions_template/models/event_question_template.py
```diff
@@ -0,0 +1,59 @@
...
+    )
+    question_id = fields.Many2one(
+        comodel_name='event.question.template.question',
+        required=True,
+        ondelete='cascade',
+    )
+    sequence = fields.Integer(default=10)
+
+
+class EventQuestionTemplate(models.Model):
```
pedrobaeza: Why putting main object at the end? It's not needed and it's more confusing.

website_event_questions_template/models/res_config.py
```diff
@@ -0,0 +1,20 @@
...
+
+from odoo import api, fields, models
+
+
+class EventMailConfigSettings(models.TransientModel):
+    _inherit = 'event.config.settings'
+
+    event_question_template_id = fields.Many2one(
+        comodel_name='event.question.template',
+        string='Questions Template',
```
pedrobaeza: String should be `Default questions template`. The problem I see with this is that if you set a default template, this template will always be put on new events, and thus questions will be filled, so if you want to put another template, you will need to previously empty the question list manually, which is a total pain.

I would directly drop this feature. It's very easy to select a template when creating an event.

website_event_questions_template/tests/test_website_event_questions_template.py
```diff
@@ -0,0 +1,124 @@
...
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo.tests.common import SavepointCase
+from odoo.exceptions import UserError
+
+
+class EventQuestionTemplate(SavepointCase):
+
+    @classmethod
+    def setUpClass(self):
```
pedrobaeza: It's not `self`, it's `cls`

website_event_questions_template/tests/test_website_event_questions_template.py
```diff
@@ -0,0 +1,124 @@
...
+                }),
+                (0, 0, {
+                    'title': 'question T2 02',
+                    'sequence': 10,
+                    'is_individual': True,
+                }),
+            ],
+        }
+        event = self.env['event.event'].create(vals)
+        with self.assertRaises(UserError), self.cr.savepoint():
```
pedrobaeza: Why do you need `self.cr.savepoint()`?

website_event_questions_template/views/website_event_questions_template_view.xml
```diff
@@ -0,0 +1,69 @@
...
+    <field name="arch" type="xml">
+        <form string="Event Question Template">
+            <sheet>
+                <group>
+                    <field name="name"/>
+                </group>
+                <group string="Questions">
+                    <field name="question_ids" nolabel="1"
+                           context="{'default_template_id': active_id}">
+                        <tree string="Questions">
```
pedrobaeza, line 30: You shouldn't define an embedded tree, but try to have a primary view inheriting from the other object view. I have 2 doubts with this method:

* Does Odoo allow to define a "cross-model" view?
* Inherited views from the main one are also used in this primary view? I'm pretty that this answer is negative.

website_event_questions_template/models/event.py
```diff
@@ -0,0 +1,42 @@
...
+        self.ensure_one()
+        vals = [(6, 0, [])]
+        if self.question_ids:
+            raise UserError(_(
+                "You can not load a template if there are defined questions"))
+        if self.event_question_template_id:
+            for question in self.event_question_template_id.question_ids:
+                vals.append((0, 0, {
```
pedrobaeza: For making this auto extensible, you should use this code:
```python
for question in self.event_question_template_id.question_ids:
    vals.append((0, 0, question._convert_to_write(question._cache)))
```
Simpler, elegant, and serves for all the new custom fields (but requires the other change I have mentioned).

pedrobaeza (changes requested):
And check Travis result after all the changes.

pedrobaeza:
Other thing: you have put in the title [9.0], but both branches mention 10.0.

website_event_questions_template/models/res_config.py
pedrobaeza: I'm remembering that this comment doesn't apply as this is loaded on demand, but any way I still think a default one is not the best option (as in v11 we will be able to put this in an onchange - Please add a comment about this in "Known issues / Roadmap").

website_event_questions_template/i18n/es.po
sergio-teruel: What app are you using to translate?

pedrobaeza: Text editor :wink:

yajo: If you're a human, you can use POEdit's *Catalog > Purge unused translations* instead 😋

rafaelbn:
Please @sergio-teruel could you review comments? Thanks

website_event_questions_template/views/website_event_questions_template_view.xml
sergio-teruel: Are you sure it's worth it?
The view that you say is does an inherit from event... if I do a primary view I have to remove this inheritance....

website_event_questions_template/models/event.py
sergio-teruel: I have an issue... the first is empty from template... mhuuu... I think to research this...

yajo: Hmm maybe with @pedrobaeza's method you would also copy reserved columns such as `id`, `create_uid`, etc.

I guess the most elegant would be to make those template models have a method that un-templates them, maybe `consolidate()` or whatever. Then you could do here:

```python
self.question_ids = self.event_question_template.question_ids.consolidate()
```

Another option would be to not have separate models. I think that's becoming the trend lately. For instance, they dropped the mass mailing templates and fell back to simply duplicating them. Have you considered this approach? You'd get all possible fields and behavior out of the box, and a simple `.copy()` would do the dirty job...

website_event_questions_template/views/website_event_questions_template_view.xml
pedrobaeza: The goal of this is to have the same inherited views as the main one without glue modules. If it's not possible, then don't do it. Do you understand me what I want to achieve?

website_event_questions_template/models/event.py
pedrobaeza: No, `_convert_to_write` doesn't pass that values. I don't get any way what is the problem right now.

sergio-teruel:
@pedrobaeza changes done..

rafaelbn (approved):
Tested functionally in runbot 👍 Thanks @sergio-teruel

website_event_questions_template/models/event.py
```diff
@@ -0,0 +1,37 @@
...
+            self.event_question_template_id.mapped('question_ids.title')
+            self.event_question_template_id.mapped(
+                'question_ids.answer_ids.name')
+            for question in self.event_question_template_id.question_ids:
+                question_vals = question._convert_to_write(question._cache)
+                question_vals['answer_ids'] = [
+                    (0, 0, x._convert_to_write(x._cache)) for x
+                    in question.answer_ids]
+                vals.append((0, 0, question_vals))
+            self.question_ids = vals
```
pedrobaeza: You should always assign this value, even if no `self.event_question_template_id`. This way, you are able to empty a template.

website_event_questions_template/models/event.py
```diff
@@ -0,0 +1,37 @@
...
+        comodel_name='event.question.template',
+        string='Questions Template',
+    )
+
+    def load_question_template(self):
+        # Onchange method is not possible due to
+        # https://github.com/odoo/odoo/issues/20163
+        self.ensure_one()
+        vals = [(6, 0, [])]
+        if self.question_ids:
```
pedrobaeza: Restrict this to 
```
if self.question_ids and self.event_question_template_id:
```
This way, you are able to empty a template.

website_event_questions_template/views/website_event_questions_template_view.xml
pedrobaeza: Is this possible then?

pedrobaeza (changes requested):

rafaelbn:
@pedrobaeza are your comments blocking? if they are light or suggestion could we just merge please?

pedrobaeza:
We should check them. @chienandalu please try what I propose.

chienandalu:
Made some experiments with what you purposed about views. The most promising way to achieve an inherited view from the inherited model was with something like this:
```python
    @api.model
    def fields_view_get(self, view_id=None, view_type='form', toolbar=False,
                        submenu=False):  # pragma: no cover
        if not view_id:
            view_id = self.env['ir.ui.view'].search([
                ('model', '=', 'event.question'),
                ('xml_id', '=',
                 'website_event_questions.view_event_question_form'),
            ], limit=1).id
        return  super(EventQuestionTemplateQuestion, self).fields_view_get(
            view_id=view_id, view_type=view_type, toolbar=toolbar,
            submenu=submenu)
```
But I'm not sure if it's going to overcomplicate inheritance...

pedrobaeza:
No, you don't need that. Why don't you simply try to create an XML view inheriting from one of the other model?

chienandalu:
Yes. I also tried that and it worked, but it broke the inheritance of the view.

pedrobaeza:
You have to add `mode="primary"`.

chienandalu:
Ok, that worked :) Anyway, there's no primary tree view for the question model, so that cannot be inherited the same way

pedrobaeza:
And do you need a tree view for templates? Isn't enough with the template name that Odoo provides by default?

chienandalu:
I mean here: https://github.com/OCA/event/pull/95#discussion-diff-148318380R43 Default tree view gives you only the name of the records.

pedrobaeza:
Agh, I see that it's embedded in the event.event form. What a pity! OK then, we should continue this way as there's no other choice. Consider to make a PR to Odoo master for extracting that tree view to one not embedded.

rafaelbn:
Could we merge then this PR @pedrobaeza ? Five months in producción and working from the 1st day :thinking:  :+1:

pedrobaeza (approved):

PR #96, opened 2017-11-09
[10.0][IMP] partner_event: Set attendee partner

sergio-teruel:
cc @Tecnativa
Please @rafaelbn @chienandalu Can you review??

### DIFF
--- a/partner_event/README.rst
+++ b/partner_event/README.rst
@@ -10,8 +10,8 @@ smart button.
 
 It also includes:
 
-* Search partners by their event registrations.
-* Search partners by number of events registered.
+* Search partners by their event attendees.
+* Search partners by number of events attendees.
 * Search partners by number of events attended.
 * Partner column is visible on registration one2many list inside the event.
 * Action in partner tree view 'More' button, to register several partners
@@ -62,12 +62,13 @@ Images
 Contributors
 ------------
 
-* Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>
+* Pedro M. Baeza <pedro.baeza@tecnativa.com>
 * Javier Iniesta <javieria@antiun.com>
-* Antonio Espinosa <antonioea@antiun.com>
+* Antonio Espinosa <antonioea@tecnativa.com>
 * Jairo Llopis <jairo.llopis@tecnativa.com>
 * Vicent Cubells <vicent.cubells@tecnativa.com>
 * David Vidal <david.vidal@tecnativa.com>
+* Rafael Blasco <rafael.blasco@tecnativa.com>
 
 Maintainer
 ----------
--- a/partner_event/__manifest__.py
+++ b/partner_event/__manifest__.py
@@ -8,7 +8,7 @@
 
 {
     'name': 'Link partner to events',
-    'version': '10.0.1.0.0',
+    'version': '10.0.2.0.0',
     'category': 'Marketing',
     'author': 'Tecnativa,'
               'Odoo Community Association (OCA)',
--- a/partner_event/models/event_registration.py
+++ b/partner_event/models/event_registration.py
@@ -11,6 +11,12 @@ from odoo import api, fields, models
 class EventRegistration(models.Model):
     _inherit = "event.registration"
 
+    attendee_partner_id = fields.Many2one(
+        comodel_name='res.partner',
+        string='Attendee Partner',
+        ondelete='restrict',
+    )
+
     def _prepare_partner(self, vals):
         return {
             'name': vals.get('name') or vals.get('email'),
@@ -20,25 +26,24 @@ class EventRegistration(models.Model):
 
     @api.model
     def create(self, vals):
-        if not vals.get('partner_id') and vals.get('email'):
-            partner_model = self.env['res.partner']
-            event_model = self.env['event.event']
-            partner_id = False
+        if not vals.get('attendee_partner_id') and vals.get('email'):
+            Partner = self.env['res.partner']
+            Event = self.env['event.event']
             # Look for a partner with that email
             email = vals.get('email').replace('%', '').replace('_', '\\_')
-            partner = partner_model.search(
-                [('email', '=ilike', email)], limit=1)
-            event = event_model.browse(vals['event_id'])
-            if partner:
-                partner_id = partner.id
-                vals['name'] = vals.get('name') or partner.name
-                vals['phone'] = vals.get('phone') or partner.phone
+            attendee_partner = Partner.search([
+                ('email', '=ilike', email)
+            ], limit=1)
+            event = Event.browse(vals['event_id'])
+            if attendee_partner:
+                vals['name'] = vals.setdefault('name', attendee_partner.name)
+                vals['phone'] = vals.setdefault(
+                    'phone', attendee_partner.phone)
             elif event.create_partner:
                 # Create partner
-                partner = partner_model.sudo().create(
+                attendee_partner = Partner.sudo().create(
                     self._prepare_partner(vals))
-                partner_id = partner.id
-            vals['partner_id'] = partner_id
+            vals['attendee_partner_id'] = attendee_partner.id
         return super(EventRegistration, self).create(vals)
 
     @api.multi
--- a/partner_event/views/event_registration_view.xml
+++ b/partner_event/views/event_registration_view.xml
@@ -4,6 +4,26 @@
      License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
 <odoo>
 
+<record id="view_event_registration_form" model="ir.ui.view">
+    <field name="model">event.registration</field>
+    <field name="inherit_id" ref="event.view_event_registration_form"/>
+    <field name="arch" type="xml">
+        <field name="name" position="after">
+            <field name="attendee_partner_id"/>
+        </field>
+    </field>
+</record>
+
+<record id="view_event_registration_tree" model="ir.ui.view">
+    <field name="model">event.registration</field>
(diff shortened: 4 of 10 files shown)


partner_event/models/event_registration.py
```diff
@@ -20,25 +26,24 @@ def _prepare_partner(self, vals):
...
     @api.model
     def create(self, vals):
-        if not vals.get('partner_id') and vals.get('email'):
-            partner_model = self.env['res.partner']
-            event_model = self.env['event.event']
-            partner_id = False
+        if not vals.get('attendee_partner_id') and vals.get('email'):
+            Partner = self.env['res.partner']
+            Event = self.env['event.event']
+            attendee_partner_id = False
```
chienandalu: Is this line necesary?

partner_event/models/event_registration.py
```diff
@@ -20,25 +26,24 @@ def _prepare_partner(self, vals):
...
-            if partner:
-                partner_id = partner.id
-                vals['name'] = vals.get('name') or partner.name
-                vals['phone'] = vals.get('phone') or partner.phone
+            attendee_partner = Partner.search([
+                ('email', '=ilike', email)
+            ], limit=1)
+            event = Event.browse(vals['event_id'])
+            if attendee_partner:
+                vals['name'] = vals.get('name') or attendee_partner.name
```
chienandalu: Maybe force here the attendee_partner name?

chienandalu (changes requested):
Tested on runbot ok.

It'd be nice to replicate also the onchange here with the attendee partner: https://github.com/odoo/odoo/blob/10.0/addons/event/models/event.py#L381

partner_event/models/event_registration.py
sergio-teruel: I have doubts... If in my second register with the same email I don't know what to do.
Do we want the new name or the save name in partner record? ping @rafaelbn

sergio-teruel:
@chienandalu Changes done

chienandalu (changes requested):
Maybe you have to tweak the tests for the new changes

rafaelbn (approved):
functionally tested it works like a charm! 👍 

Please update:

- Version
- Migration
- Contributors (tecnativa + add me)
- Travis greeen

Thanks

rafaelbn:
@sergio-teruel 



functionally tested it works like a charm! 👍

Please update:

    Version
    Migration
    Contributors (tecnativa + add me)
    Travis greeen

Thanks

partner_event/models/event_registration.py
```diff
@@ -20,25 +26,23 @@ def _prepare_partner(self, vals):
...
+                vals['phone'] = vals.get('phone') or attendee_partner.phone
```
yajo: Use `vals.setdefault()` instead.

partner_event/models/event_registration.py
```diff
@@ -11,6 +11,12 @@
 class EventRegistration(models.Model):
     _inherit = "event.registration"
 
+    attendee_partner_id = fields.Many2one(
+        comodel_name='res.partner',
+        string='Attendee Partner',
```
yajo, line 16: 🤔 I don't quite understand what's the field for?

chienandalu: There was no `partner_id` in `event.registration` model before 9.0 in wich Odoo introduced. The problem is that Odoo's use is for billing purposes whereas that `event_partner` uses it to create or link each attendee to a different contact. So this field is to split those different behaviours.

yajo: And what would happen if we use `partner_id` for this latter purpose and use the partner's billing data for bills?

I'd also thank some docs on this matter in the readme/code.

chienandalu (approved):

partner_event/models/event_registration.py
sergio-teruel: I think that this question is decided by Odoo :smile:

partner_event/__manifest__.py
```diff
@@ -8,7 +8,7 @@
 
 {
     'name': 'Link partner to events',
-    'version': '10.0.1.0.0',
+    'version': '10.0.1.0.1',
```
yajo: Hmm this should be `10.0.2.0.0`, since it means running a DB migration and is not completely backwards-compatible, right? 🤔

partner_event/README.rst
```diff
@@ -62,12 +62,13 @@ Images
...
 
-* Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>
+* Pedro M. Baeza <pedro.baeza@tecnativa.com>
 * Javier Iniesta <javieria@antiun.com>
-* Antonio Espinosa <antonioea@antiun.com>
+* Antonio Espinosa <antonioea@tecnativa.com>
 * Jairo Llopis <jairo.llopis@tecnativa.com>
 * Vicent Cubells <vicent.cubells@tecnativa.com>
 * David Vidal <david.vidal@tecnativa.com>
+* Rafael Blasco <rafael.blasco@tecnativa.com>
```
yajo, line 71: @sergio-teruel you can add yourself here if you want 😊

yajo (changes requested):

rafaelbn:
@sergio-teruel please review if we can apply this improvement here: https://github.com/OCA/event/pull/97

rafaelbn:
@Yajo change done, please review again thanks

yajo (approved):

partner_event/views/event_registration_view.xml
```diff
@@ -4,6 +4,26 @@
      License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
 <odoo>
 
+<record id="view_event_registration_form" model="ir.ui.view">
+    <field name="model">event.registration</field>
+    <field name="inherit_id" ref="event.view_event_registration_form"/>
+    <field name="arch" type="xml">
+        <field name="name" position="after">
+            <field name="attendee_partner_id"/>
```
gustavovalverde, line 12: I can't understand why the `attendee_partner_id` is taking more relevance than the `partner_id` with this PR, I feel like this changes a working behavior and it's giving more relevance to a field that actually does not keep all the partner's logic.

I do understand that the attendee_partner_id could be different from the partner_id. But if the partner_id is the one paying, why would the attendee_partner_id to be more relevant?

gustavovalverde: I also feel this causes 'billing inconsistency' for a company that their core business and billing is based on events, as a partner shouldn't have a billing that differs from their event registration.

yelizariev: @gustavovalverde your concern is that ``attendee_partner_id`` should not be introduced?
What is the case when attendee shall be the same as person  that paid the ticket? It limits the sale, doesn't?

pedrobaeza: I don't see any problem, as they are different fields.

gustavovalverde: I'm no against of having the `attendee_partner_id` , but just of making this field *more relevant* (in a functional way) than the `partner_id` makes it a bit harder for analysis of what's going on.

For example: 
image

It's more common a company sending their employees (partner_childs_id), than a company serving as a middleman (this could be around a 10%), and only in the latter case the `attendee_partner_id` is more relevant.

In v8 it was taking the events count from the partner (which makes sense as an SO can only have 1 partner), now we saw the analytics and it was distributing the event count to the attendees.

Summarizing: I'm not saying to remove this field, It's just that for most uses cases that I've worked on in the events and education industry this field is just not more relevant than the partner_id.

pedrobaeza: More relevant in which part exactly? For mass mailings, it's more relevant to send the event information to the contact than to the company.

gustavovalverde: For mailing lists I already have the Attendee Name, their Phone and Email. Having a partner for the attendee is not needed for that specific case.

This could be even marked as a "taste" thing. For my clients, I just removed the field from the view in the registration (so it wouldn't affect the event count). I just wanted to make the __heads up__.

I'll add this to my work pipe, to consider a PR here, where this behavior can be 'customized':
- Relate event based on Attendee Partner
- Relate event based on Partner

rafaelbn: @gustavovalverde 

>  can't understand why the attendee_partner_id is taking more relevance than the partner_id with this PR, I feel like this changes a working behavior and it's giving more relevance to a field that actually does not keep all the partner's logic.

> I do understand that the attendee_partner_id could be different from the partner_id. But if the partner_id is the one paying, why would the attendee_partner_id to be more relevant?

Why more relevance? This doesn't mean more relevance mean each one has relevance.

- One is the the res.partner who pay
- Two is the res.partner who attend
- One and two could be equal or not
- One has relevance for Sale, Invoice purpose
- Second has relevance for Event analysis purpose

> For mailing lists I already have the Attendee Name, their Phone and Email. Having a partner for the attendee is not needed for that specific case

Yes, is needed. Is basic. Because your Partners (companies and contacts) are your customer you manage this in you CRM and this module make this, relate you Partners (CRM) with you Event management.

@gustavovalverde if you don't understand or need it, just don't use this module because if understand you, what you expect is what Odoo makes by default. This module is to improve and add functionality.

Please go to Odoo  (without this module) and let me know all your res.partner that has attendee to one single event. You cannot. You can only tell my which partners has paid and event. If you want to know which partners has attended to an event then you need this module.

Note also that you could have 1 sale order and 1 partner who paid but 10 different attendees (which are also partners)

Hope this helps

PR #99, opened 2017-12-31
[NEW] event_sale_confirmation: hide button confirm in some cases

yelizariev:
Module hides ``[Confirm]`` button when ticket is not free and user doesn't have access level ``Events: Manager``.

### DIFF
--- a/event_sale_confirmation/__manifest__.py
+++ b/event_sale_confirmation/__manifest__.py
@@ -1,0 +1,23 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017,2018 IT-Projects LLC - Ivan Yelizariev
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
+{
+    "name": "Restricted access to confirm unpaid tickets",
+    "summary": "Only event manager can confirm unpaid tickets",
+    "version": "10.0.1.0.0",
+    "category": "Event Management",
+    "website": "https://github.com/OCA/event",
+    "author": "IT-Projects LLC, Odoo Community Association (OCA)",
+    "license": "LGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        "event_sale",
+    ],
+    "data": [
+        "views/event_registration_view.xml",
+    ],
+    "demo": [
+        "views/tour_views.xml",
+    ],
+}
--- a/event_sale_confirmation/models/event_registration.py
+++ b/event_sale_confirmation/models/event_registration.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017,2018 IT-Projects LLC - Ivan Yelizariev
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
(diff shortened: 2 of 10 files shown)


event_sale_confirmation/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+    "name": "Restricted access to confirm unpaid tickets",
+    "summary": "Only event manager can confirm unpaid tickets",
+    "version": "10.0.1.0.0",
+    "category": "Event Management",
+    "website": "https://github.com/OCA/event",
+    "author": "IT-Projects LLC, Odoo Community Association (OCA)",
+    "license": "LGPL-3",
+    "application": False,
+    "installable": True,
+    "external_dependencies": {
```
pedrobaeza: Don't put empty keys

event_sale_confirmation/models/event_registration.py
```diff
@@ -0,0 +1,18 @@
...
+# Copyright 2017,2018 IT-Projects LLC - Ivan Yelizariev
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
+from odoo import models, fields, api
+
+
+class EventRegistration(models.Model):
+
+    _inherit = 'event.registration'
+
+    event_ticket_price = fields.Float(related='event_ticket_id.price')
```
pedrobaeza: readonly=True for avoiding rewritings each time.

pedrobaeza (changes requested):

event_sale_confirmation/models/event_registration.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class EventRegistration(models.Model):
+
+    _inherit = 'event.registration'
+
+    event_ticket_price = fields.Float(
+        related='event_ticket_id.price',
+        readonly=True)
+    user_is_manager = fields.Boolean(compute='_compute_user_is_manager')
```
chienandalu, line 14: I'm not sure if something like `can_confirm` would be a more appropiate name for this field

chienandalu (approved):
Looks good to me :+1:

chienandalu (commented):
It shouldn't be hard to do some tests for coverage

event_sale_confirmation/models/event_registration.py
yelizariev: It's not so. The user can confirm when he is a manager OR ticket is free

yelizariev:
What shall I do with this warning;
```

event_sale_confirmation/static/src/js/tour.js:62: [W7903(javascript-lint), ] Unexpected console statement. [Error/no-console]
event_sale_confirmation/static/src/js/tour.js:77: [W7903(javascript-lint), ] Unexpected console statement. [Error/no-console]
```

``console.log`` is an essential part of tour tests

yelizariev:
Anyway, travis is green

event_sale_confirmation/models/event_registration.py
chienandalu: True that :+1:

PR #101, opened 2018-02-16
[ADD] website_event_snippet_calendar: Snippet for event calendar and list

yajo:
Allows the user to display a calendar combined with a list of events.

Includes controllers for such information to be bundled into any page through an `<iframe>`.

Linter will be affected by https://github.com/OCA/pylint-odoo/pull/189

@Tecnativa


website_event_snippet_calendar/templates/snippets.xml
```diff
@@ -0,0 +1,84 @@
...
+        <xpath expr="//*[@id='snippet_feature']//*[hasclass('o_panel_body')]">
+            <t t-snippet="website_event_snippet_calendar.s_event_calendar_list"
+               t-thumbnail="/website_event_snippet_calendar/static/description/icon.png"/>
+        </xpath>
+    </template>
+
+    <!-- Add snippets options -->
+    <template id="snippet_options" inherit_id="website.snippet_options">
+        <xpath expr=".">
+            <!-- Allow to add or remove events from the list -->
```
chienandalu, line 70: Please, document in README :wink:

chienandalu (approved):
Tested in runbot. Just a non blocking comment

website_event_snippet_calendar/templates/snippets.xml
yajo: Done in https://github.com/OCA/event/pull/101/commits/3b6395983ad82d86d8ffe98950de222873190d39

luismontalba (approved):
Tested on runbot

yajo:
This one should be ready to squash and merge.

website_event_snippet_calendar/README.rst
```diff
@@ -0,0 +1,68 @@
...
+help us smash it by providing detailed and welcomed feedback.
+
+Credits
+=======
+
+Contributors
+------------
+
+* `Tecnativa <https://www.tecnativa.com>`_:
+  * Jairo Llopis <jairo.llopis@tecnativa.com>
```
pedrobaeza: Add an extra line as this is not correctly rendered

website_event_snippet_calendar/controllers/main.py
```diff
@@ -0,0 +1,72 @@
...
+    def days_with_events(self, start, end):
+        """Let visitors know when are there going to be any events.
+
+        :param start string:
+            Search events from that date.
+
+        :param end string:
+            Search events until that date.
+        """
+        events = request.env["event.event"].search([
```
pedrobaeza, line 22: What about published flag?

website_event_snippet_calendar/controllers/main.py
```diff
@@ -0,0 +1,72 @@
...
+
+        :param day string:
+            Date in a string. If ``None``, we'll search for upcoming events
+            from today up to specified limit.
+
+        :param limit int:
+            How many results to return.
+        """
+        ref = day or Date.to_string(date.today())
+        domain = [
```
pedrobaeza, line 52: The same about published ones

pedrobaeza (changes requested):
Please make the icon square, or it will be deformed on module list

website_event_snippet_calendar/controllers/main.py
yajo: It's not necessary because the permission rules already force that.

This way, authorized users can also see unpublished events. That's how upstream website events list works too.

website_event_snippet_calendar/README.rst
yajo: Done

pedrobaeza: It should be before, not after

website_event_snippet_calendar/controllers/main.py
pedrobaeza: OK

yajo:
The icon proportion is respected. I remember what you say happened before, but possibly they fixed it since v10+:

captura de pantalla de 2018-04-02 08-18-23
captura de pantalla de 2018-04-02 08-18-32

It seems there are no more comments pending, right?

pedrobaeza (approved):

PR #103, opened 2018-02-19
[9.0][ADD] crm_event_registration_partner_unique

chienandalu:
This module prevents duplicated partners in an event with *Forbid duplicates* enabled when partners in that event are merged.

Autoinstalls when `crm` is installed along with `event_registration_partner_unique`.

cc @Tecnativa

### DIFF
--- a/crm_event_registration_partner_unique/__openerp__.py
+++ b/crm_event_registration_partner_unique/__openerp__.py
@@ -1,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Unique Partner per Event and CRM",
+    "summary": "Avoids duplicates in events when partners are merged",
+    "version": "9.0.1.0.0",
+    "category": "Events",
+    "website": "http://github.com/OCA/event",
+    "author": "Tecnativa, "
+              "Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "auto_install": True,
+    "depends": [
+        "event_registration_partner_unique",
+        "crm",
+    ],
+    "data": [
+    ],
+}
--- a/crm_event_registration_partner_unique/README.rst
+++ b/crm_event_registration_partner_unique/README.rst
@@ -1,0 +1,53 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+================================
+CRM and Unique Partner per Event
+================================
+
+This module prevents duplicated partners in an event with *Forbid duplicates*
+enabled when partners in that event are merged.
+
+Autoinstalls when `crm` is installed along with
+`event_registration_partner_unique`.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/199/9.0
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/event/issues>`_. In case of trouble, please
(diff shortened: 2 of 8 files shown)


rafaelbn (approved):
tested in runbot @chienandalu :+1: 

Very important this module @Yajo

pedrobaeza:
The name is totally incorrect. It should be `crm_deduplicate_event_registration_partner_unique`. I see that the original module is also badly implemented, as it can be done with an SQL constraint.

chienandalu:
@pedrobaeza But the original module gives the option to have then event with unique partners or not, so an sql contraint wouldn't apply, isn't it?

pedrobaeza:
You're right, but I think it has no sense to not check that one in any list. For what do you want duplicates?

crm_event_registration_partner_unique/__openerp__.py
```diff
@@ -0,0 +1,22 @@
...
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "auto_install": True,
+    "depends": [
+        "event_registration_partner_unique",
+        "crm",
+    ],
+    "data": [
+    ],
```
yajo, line 21: :fire:

yajo (commented):
Hmm I think this addon is wrong. Let me explain:

I did this one test:

1. Install `website_event_sale`
2. Register an attendee to a paid event.
3. **When the attendee is registered, the `partner_id` is *Public User*.**
4. Confirm the SO.
5. **Now the `partner_id` becomes the *billing partner* I used when registering**

So, from the above, In Odoo v9+, the `partner_id` field identifies the *billing partner*, not the attendee itself. Thus, duplicated partners in `partner_id` field from `event.registration` model are a necessity, because you can either:

- Have several unidentified attendees, which are all *Public User*.
- Have several paid attendees from the same *billing partner*.

The patch should go then to `partner_event` which defines the `attendee_partner_id` field as you can see here: https://github.com/OCA/event/blob/5843da351336650eda9bda90abec2eec799ea491/partner_event/models/event_registration.py#L17-L21

The main problem is that the code above is from the 10.0 branch. In v9, this addon uses the `partner_id` for such purpose, which IMHO must have been a misunderstanding on the purpose of the field at module migration time, since that changed from how it was at v8 (and Odoo notified nobody, of course). I'm not sure if at this time backporting the v10 implementation would produce more problems than benefits, but indeed v9 implementation seems to be wrong.

In any case, on the issue at hand, a quick search revealed that [Postgres supports exclusion constraints](https://stackoverflow.com/a/48200472/1468388), which can be used to do the check via SQL and quickly fix all of this.

pedrobaeza:
@Yajo this is for v9. The behavior you describe is for v10, isn't it?

yajo:
That's what I was thinking at first, but then I did the test in this very PR runbot and it's the same. The behavior didn't change for v10, but for v9; it was us that didn't notice until then. Just do the test and see. :confused:

pedrobaeza:
Any way, this module is used not for registering via website, but for introducing attendees via backend. I think it's still useful in that context.

chienandalu:
@Yajo Thing is these modules don't depend on partner_event in any way and i'd say it's not used in the manner you expose :slightly_smiling_face:

yajo:
The point I'm trying to say is that if you forbid duplicates, you will break Odoo normal operation. For instance, a single person that browses the website and subscribes 2 attendees for the same event (no matter if it is paid or not) would duplicate the Public User in such registrations, and thus would get a 500 error.

chienandalu:
But maybe that's because the target of the unique partner events isn't that one...

rafaelbn:
@chienandalu @Yajo please could we go ahead with this PR please? Thanks! :+1:

yajo (approved):
OK, the problem in this PR is not exactly this new addon, but the parent addon where it is based on: `event_registration_partner_unique`, which implements an idea that is wrong per se. It's an addon that should have never existed in v9, where now the `partner_id` is not the attendee but the payer.

In any case, in deployments where this concept somehow works, indeed this addon fixes a problem and is well developed, so I will merge now.

Let's just make sure we don't migrate any of these to v10+.

OTOH I imagine the problem that this is fixing will happen in `event_partner`, although in the right field, but I guess tomorrow we'll fix tomorrow's problems.

PR #104, opened 2018-02-20
[FIX] partner_event: Respect attendee language

yajo:
- When attendees were registered from the website and the event was configured to create a partner, the partner did not have the user's language.
- When printing the registration badge, it was printed in linked billing partner or current user's language. Now it is printed in the attendee language if available.
- When writing in the chatter, attendee partner is suggested as follower.

@Tecnativa

### DIFF
--- a/partner_event/tests/test_event_registration.py
+++ b/partner_event/tests/test_event_registration.py
@@ -10,6 +10,7 @@ from psycopg2 import IntegrityError
 from datetime import datetime, timedelta
 from odoo.tests import common
 from odoo import fields
+from ..hooks import LANG_NEW, LANG_OLD, uninstall_hook
 
 
 class TestEventRegistration(common.SavepointCase):
@@ -39,6 +40,11 @@ class TestEventRegistration(common.SavepointCase):
         cls.registration_02 = registration_model.create({
             'email': 'email02@test.com', 'event_id': cls.event_0.id,
             'name': 'Test Registration 02', 'phone': '254728911'})
+        # Need a 2nd lang to do some lang-specific tests
+        wiz = cls.env["base.language.install"].create({
+            "lang": "it_IT",
+        })
+        wiz.lang_install()
 
     def test_create(self):
         self.assertEqual(self.partner_01.name, self.registration_01.name)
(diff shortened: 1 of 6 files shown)


rafaelbn:
Hi @Yajo ,

It doesn't work

This is the partner:
- http://3318001-104-4e438c.runbot2.odoo-community.org/web?debug#id=46&view_type=form&model=res.partner
Partner is correctly created with the language of the website! Great! :+1: :tada: 

This is the message (still in english):
- http://3318001-104-4e438c.runbot2.odoo-community.org/web?debug#id=354&view_type=form&model=mail.message&menu_id=81&action=103

Maybe it is because an new one BUG in Odoo side? As you can check there is no recipient in the message....

Thank you!

yajo:
The issue you say is Odoo's bug, notified internally to them as OPW-817260.

pedrobaeza (approved):

chienandalu (commented):
Coverage fell dramatically :confused: 

`27.27% of diff hit (target 95.13%)`

yajo:
Noo, the diff hit is how many introduced lines are tested (codecov/patch).
The project coverage (codecov/project) falls only -1.24%.
In any case, you're right, this needs tests. :wink:

pedrobaeza:
Travis is failing

rafaelbn (approved):
Tested functionally :+1:  thanks!

But @Yajo please add teste and review travis :red_circle:

rafaelbn:
@Yajo please review travis here is :red_circle: and we would like to merge this in order to migrate to v11

cc @jbeficent take care here

pedrobaeza:
Shouldn't this be done on the mail template, like https://github.com/odoo/odoo/pull/24718?

partner_event/tests/test_event_registration.py
```diff
@@ -37,6 +37,8 @@ def setUpClass(cls):
         cls.registration_02 = registration_model.create({
             'email': 'email02@test.com', 'event_id': cls.event_0.id,
             'name': 'Test Registration 02', 'phone': '254728911'})
+        # Need a 2nd lang to do some lang-specific tests
+        cls.env['res.lang'].load_lang("it_IT")
```
chienandalu: Maybe you need to load `es_ES ` also

chienandalu:
@pedrobaeza Indeed it is needed to change the templates default recipient and lang settings. Otherwise it's not going to work (testet in runbot)

chienandalu (commented):
@Yajo The default settings on the templates override all this changes

image
image

yajo:
1st picture should be fixed in https://github.com/odoo/odoo/pull/24718 instead.
2nd picture should be fixed with last commit :blush:

chienandalu:
1st -> Ok, yeah, but, maybe you it could be interesting to have a default like:

```
${object.email or object.attendee_partner_id or object.parnter_id}

rafaelbn:
@chienandalu @Yajo whats going on with this PR? This has been in production for a long time!
The problem here is that we could migrate modules to v11 or v12 and miss this FIXES

PR #107, opened 2018-03-26
[IMP] event_email_reminder: Allow negative intervals

yajo:
This way, we can schedule an email after an event has finished.

@Tecnativa


pedrobaeza (approved):
Bump version number

rafaelbn (approved):

PR #108, opened 2018-03-29
[ADD] event_track_location_overlap: Control track location overlapping

yajo:
New module that allows to forbid overlapping multiple tracks in the same room at the same time... or not.

@Tecnativa


pedrobaeza (changes requested):
This won't work on track creation, as the search won't return yet the new record. You can't put outside the constraint method (`_check_location_overlap`) the check, as you need data in `self` for comparing.

yajo:
But... tests pass, and they do creation to test.

pedrobaeza:
Please try on your local.

pedrobaeza (approved):

rafaelbn:
Functionally I don't thinks this is correct, check screenshoot:



- In this case I created a event in 2018-04-06
- But the track is in other date which the location is empty and available

rafaelbn (changes requested):
Functionally tested I comment a used case with an error

yajo:
In Odoo, the event start and end dates are manually saved into the `event.event` object, and displayed when calling its `.display_name` field. That's what you are seeing in the exception message (the part you highlighted).

However, nothing forbids a user from saving an event track in a date that falls outside the event dates.

So, possibly that's the problem you are facing. Could you check please? Are you sure both tracks were not overlapping?

Thanks

rafaelbn:
> So, possibly that's the problem you are facing. Could you check please? Are you sure both tracks were not overlapping?

Yes, I'm sure

yajo:
Sorry @rafaelbn I cannot reproduce your failure. I rebuilt the runbot. Could you try again and provide better STR please?

pedrobaeza:
@rafaelbn update your review

chienandalu (commented):
The migration to v12 from this is already merged. Shouldn't this one be so?

pedrobaeza:
Yes, let's move on.

PR #109, opened 2018-03-29
[FIX] event_registration_mass_mailing: Fix integration tests

yajo:
Testing after installing mass_mailing_list_dynamic produces:

    IntegrityError: null value in column "sync_domain" violates not-null constraint

@Tecnativa


pedrobaeza (approved):
Fast-tracking this

PR #111, opened 2018-05-23
[FIX] partner_event: Update event.registration records without permissions

yajo:
If a given user has no permission to access `event.registration` records, he becomes unable to update some `res.partner` fields.

Now that process uses `.sudo()` and is tested to work fine.

@Tecnativa

### DIFF
--- a/partner_event/models/res_partner.py
+++ b/partner_event/models/res_partner.py
@@ -17,6 +17,7 @@ class ResPartner(models.Model):
     event_count = fields.Integer(
         string='Events',
         compute='_compute_event_count',
+        compute_sudo=True,
         help="Count of events with confirmed registrations.",
     )
     registration_count = fields.Integer(
@@ -37,7 +38,7 @@ class ResPartner(models.Model):
     def _compute_event_count(self):
         for partner in self:
             partner.event_count = len(
-                self.env["event.registration"].sudo().search([
+                self.env["event.registration"].search([
                     ("attendee_partner_id", "child_of", partner.id),
                     ("state", "not in", ("cancel", "draft")),
                 ]).mapped("event_id"))
--- a/partner_event/models/res_partner.py
+++ b/partner_event/models/res_partner.py
@@ -37,7 +37,7 @@ class ResPartner(models.Model):
     def _compute_event_count(self):
         for partner in self:
             partner.event_count = len(
-                self.env["event.registration"].search([
+                self.env["event.registration"].sudo().search([
                     ("attendee_partner_id", "child_of", partner.id),
                     ("state", "not in", ("cancel", "draft")),
                 ]).mapped("event_id"))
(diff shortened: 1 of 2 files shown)


yelizariev:
I guess it’s related to this commit https://github.com/odoo/odoo/commit/19878d011f8abdc737b7b702c4c79632670d4b78#
You may reference to it in your commit message

yajo:
Not sure it's related, actually this addon should fail as it was designed...

rafaelbn (approved):
Thanks :+1:

partner_event/models/res_partner.py
```diff
@@ -37,7 +37,7 @@ def _compute_registration_count(self):
     def _compute_event_count(self):
         for partner in self:
             partner.event_count = len(
-                self.env["event.registration"].search([
+                self.env["event.registration"].sudo().search([
```
pedrobaeza: Please take the occasion to rewrite this to use read_group for performance reasons.

pedrobaeza (changes requested):

rafaelbn (changes requested):
Sorry I need clarification:

Model: `event.registration` has:
- Odoo core field '`partner_id`' related with `res.partner` . This field is the contact NOT the attendee
- This module partner_event ADD a new field '`attendee_partner_id`' related with res.partners. This field is the attende (and diferent of the contact)

Which is the problem? I will try to reproduce. I guess the commit @yelizariev said is related to the partner_id

Thanks

partner_event/models/res_partner.py
yajo: This method is expected to be called only from form view (where this computed field is shown), so this doesn't seem a very critical performance point at first sight. Can't we leave that for the future?

pedrobaeza: Not really. It's prefetched also on other situations. It's very simple, so please do it.

yajo:
@rafaelbn the problem happens when you enter a res.partner that has some linked event.registration info, and you don't have permission to use event.registration.

pedrobaeza:
@Yajo then the solution is not this one, but put groups on the view for not being loaded.

rafaelbn:
@Yajo OK. I still don't know the steps to reproduce the issue to test de solution. Anyway then the solution is using groups and not `sudo` so are you going to change this PR? Thank you!

yajo:
I'm still not sure that's the solution. Some integration tests fail without this patch, and those render no buttons, act only at ORM level... :thinking:

pedrobaeza:
Can you tell me which one fails?

chienandalu (commented):
What's the state of this?

yajo:
According to my tests, this is needed and ready to merge.

There's one improvement that can be done, but it's not this PR main purpose and can be done later if needed. I can do it when I have time, but I don't consider it a blocking thing.

pedrobaeza:
This comment https://github.com/OCA/event/pull/111#discussion_r190867732 is still unattended, and better to use `compute_sudo=True` instead on the definition of the field.

tarteo (approved):

pedrobaeza:
I still think is better to use `compute_sudo=True`

pedrobaeza (approved):

PR #116, opened 2018-08-29
10.0 Add simple_poll module.

ntodorova:
This module makes it possible to create simple inquiries (polls) and
distribute them using a custom link. The participants can then select
an option/s and submit their answers. The results of the inquiry are nicely
summarized in the link view.

This work is part of a master thesis written by Nikolina Todorova.
The master thesis contains further information, screenshots and the conceptual
background and can be downloaded here https://www.initos.com/forschung/publikationen/year/2018/

### DIFF
--- a/simple_poll/README.md
+++ b/simple_poll/README.md
@@ -1,0 +1,87 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl
+   :alt: License: AGPL-3
+
+===============================================
+Simple Poll
+===============================================
+
+This module makes it possible to create simple inquiries (polls) and
+distribute them using a custom link. The participants can then select
+an option/s and submit their answers. The results of the inquiry are nicely
+summarised in the link view.
+
+This work is part of a master thesis written by Nikolina Todorova.
+The master thesis contains further information, screenshots and the conceptual
+background and can be downloaded here https://www.initos.com/forschung/publikationen/year/2018/
+
+Usage
+=====
+
+To create a poll and distribute it, follow the steps:
+
+* Go ot "Polls/Polls" menu
+* Press the "Create" button
+* Fill "Title", which will be your inquire question
+* Choose one of the options
+  ** "Simple Text Question" - to fill option with arbitrary text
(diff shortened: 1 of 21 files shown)


pedrobaeza:
Why here and not in OCA/survey?

pedrobaeza:
And why not using survey module?

ntodorova:
This module was developed in relation to enhancement of the event module. Which will be more focused on team events. A PR is planned for that too, that`s why the PR was created here.

The idea of the "simple_poll" module is to be able to create one question inquiries as fast and easy as possible. The key word here is "simplicity". That`s why , the survey module was explored and use for example during the development of the simple poll. From my point of view this is a great tool for creating big questionnaires and distribute them, but it is too heavy, include multiple steps and it is not practical for inquiries from "doodle" type (single question inquiry).  What do you think?

pedrobaeza:
Thanks for the explanations. Are you aware that events include simple questions like this poll to be filled when you register?

If you still want to have this, you should move this module to OCA/survey, and have the dependent module (`simple_poll_event` or similar name) here.

ntodorova:
Yes the "Website Event Question" ( website_event_question ) module was also considered. But this
module  provides functionality for  website  event
(published event) and the context here is related to not published, team events.  Additionally we want to be able to send polls independent from events. But the structure of the module was also useful and taken into account.

Ok. The name will be changed to "simple_poll_event"

pedrobaeza:
I'm not talking about renaming, but putting in OCA/survey the base module (this one), and here the other that links to events.

Any way, why do you depend on website for this module?

Other thing in a fast sight: README should be name with .rst, not .md, and check correct syntax

ntodorova:
Ok, then I will move the PR to the survey repo.
The dependency is needed because we use a website page for the survey where the participants can choose question options. 

Could you also please give me a reference for the README correct syntax?

pedrobaeza:
> The dependency is needed because we use a website page for the survey where the participants can choose question options.

On v11, you won't need `website` module for that. Put it on the roadmap section of the README so that those that migrate the module be able to know it.

ntodorova:
The PR is closed, because it was moved to the "survey" repo.
See https://github.com/OCA/survey/pull/26

PR #118, opened 2018-09-18
[MOD] cleanup travis and add makepot to create pot files

lfreeke:
I notice there weren't any pot files in this repository. This mean you cannot use weblate to translate. 
I cleaned up the travis file and added MAKEPOT="1"


pedrobaeza:
Fast-tracking this

PR #124, opened 2018-10-30
[11.0][MIG] event_registration_partner_unique: Migration to 11.0

cristinamartinrod:
This module extends the functionality of events to support forbidding to duplicate the same partner in an event.

cc @Tecnativa


chienandalu (changes requested):
Shouldn't this module depend from `partner_event`. In version `10.0` the partner field use was split in that module into the standard one (`partner_id`) is used to register the partner that registers the attendees and the  `attendee_partner_id` wich is the one related to the attendee itself.

The reason is that, for instance, with `event_sale` or `website_event_sale` you register several attendees at once with the same `partner_id`. Keeping this module as it is, it would force the user to make an order for each attendant, wich makes no sense.

Also keep in mind this sticky module as well: https://github.com/OCA/event/tree/9.0/crm_event_registration_partner_unique that avoids to end with duplicates in the event when two different partners are merged. Maybe we could implement it in a different way in this version to get rid of it.

event_registration_partner_unique/models/event.py
```diff
@@ -0,0 +1,48 @@
...
+                    registrations=dupes,
+                )
+
+    @api.multi
+    def _duplicate_search_domain(self):
+        """What to look for when searching duplicates."""
+        return [
+            ("id", "!=", self.id),
+            ("event_id", "=", self.event_id.id),
+            ("partner_id", "=", self.partner_id.id),
```
yajo: As @chienandalu said, this must be a different field. Forbidding duplicates in `partner_id`, which is the _billing partner_, would make the addon not usable.

yajo (changes requested):

event_registration_partner_unique/models/event.py
cristinamartinrod: 👍🏻 It sounds logical so I'll change it.

pedrobaeza: You need to depend now on `partner_event` module and make the restriction over `attendee_partner_id` field.

cristinamartinrod: Done!

yajo (approved):

pedrobaeza (approved):

Issue #91, opened 2017-10-03
Migration to version 11.0


pedrobaeza:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0

# Modules to migrate

- [x] crm_lead_to_event_registration - By @cristinamartinrod - #117 
- [ ] event_contact
- [ ] event_mail (only migration script in OpenUpgrade is needed; see https://github.com/OCA/event/issues/91#issuecomment-388356633)
- [ ] event_project
- [x] event_registration_cancel_reason - By @cristinamartinrod - #120 
- [x] event_registration_mass_mailing - By @chienandalu - #114 
- [x] event_registration_partner_unique - By @cristinamartinrod - #124 
- [ ] event_registration_multi_qty
- [x] event_session - By @ntsirintanis - #127 
- [x] partner_event - By @hveficent - #110 
- [x] website_event_excerpt_img - By @tarteo - #100
- [x] website_event_filter_selector - By @tarteo - #112 
- [x] website_event_require_login - By @emagdalenaC2i - #142 
- [ ] website_event_share
- [ ] website_event_type_description

rafaelbn:
@pedrobaeza , @Yajo comment us that event_mail in v11 is included in CORE https://github.com/odoo/odoo/blob/6c6c4b89a077d2345425771c854ac608ff4f136d/addons/event/models/event_mail.py#L18-L41 . Then we should take care on this for Openupgrade but module should be deprecated. Isn't it? cc @chienandalu

chienandalu:
Yeah, I just tested it and seems to have same funcionality, although depending on event.type.

AaronHForgeFlow:
actually, the author of #110 is @hveficent not me :)

pedrobaeza:
Sorry, I mix the `h` letter, as you use only your initials. Changing it.

emagdalenaC2i:
Please add to the list crm_lead_to_event_registration - By @cristinamartinrod - #117 that is already migrated and merged

emagdalenaC2i:
website_event_require_login #142

emagdalenaC2i:
@pedrobaeza Please, add website_event_require_login #142 here and you can delete the PR number in #121

pedrobaeza:
OK, done

PR #125, opened 2018-11-20
[11.0] [MIG] event_mail

ntsirintanis:
Migration of event_mail module from 10.0 to 11.0


ntsirintanis:
@NL66278 @lfreeke

event_mail/__manifest__.py
```diff
@@ -1,12 +1,12 @@
...
 # © 2016 Sergio Teruel <sergio.teruel@tecnativa.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 {
     'name': 'Event Mail',
     'summary': 'Mail settings in events',
-    'version': '10.0.1.0.0',
+    'version': '11.0.1.0.0',
     'author': 'Tecnativa, '
+              'Therp BV, '
```
pedrobaeza, line 9: I don't think the migration grants you co-authorship. You should be in contributors of course.

pedrobaeza (changes requested):

pedrobaeza:
What about this comment saying that this is not needed: https://github.com/OCA/event/issues/91#issuecomment-388356633

PR #127, opened 2018-11-28
[11.0] [MIG] event_session

ntsirintanis:
Migrated here is only the logic of 10.0 event.session module that defines sessions for events. Aspects like separate session registration, seat availability per session, mail notifications per session that were included in event.session 10.0 are now omitted, because they are not really essential for defining a session, i.e., these are extra functionalities beyond the scope of this module.
 More logic (or reintroduction of omitted aspects) may be implemented separately via a different module that depends on event.session 11.0. @NL66278 @lfreeke


event_session/models/event.py
```diff
@@ -0,0 +1,46 @@
...
+        comodel_name='event.session',
+        inverse_name='event_id',
+        string='Sessions',
+    )
+    sessions_count = fields.Integer(
+        compute='_compute_sessions_count',
+        string='Total event sessions',
+        store=True,
+    )
+    seats_expected = fields.Integer(store=True)
```
daramousk: Is the `store=True` needed here? Since this is not a computed field.

event_session/__init__.py
```diff
@@ -0,0 +1,3 @@
+from . import models
```
NL66278, line 1: Still need the license line.

event_session/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2017 David Vidal<david.vidal@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Event Sessions',
+    'version': '11.0.1.0.0',
+    'author': 'Tecnativa, '
+              'Odoo Community Association (OCA)',
+    "license": "AGPL-3",
+    'website': 'https://odoo-community.org/',
```
NL66278: https://github.com/oca/event.git

event_session/models/__init__.py
```diff
@@ -0,0 +1,3 @@
+
```
NL66278: license line

event_session/models/event.py
```diff
@@ -0,0 +1,46 @@
+# Copyright 2017 David Vidal<david.vidal@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
```
NL66278: And license lines should be https now also.

event_session/models/event.py
NL66278: No need any more for seats_expected and seats_available expected (nor for the corresponding functions).

event_session/models/event_session.py
```diff
@@ -0,0 +1,141 @@
...
+    company_id = fields.Many2one(
+        comodel_name='res.company',
+        related='event_id.company_id',
+        store=True,
+    )
+    event_id = fields.Many2one(
+        comodel_name='event.event',
+        string='Event',
+    )
+    seats_min = fields.Integer(
```
NL66278, line 30: Do away with the seats stuff

event_session/tests/test_session.py
```diff
@@ -0,0 +1,154 @@
...
+# Copyright 2017 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl-3.0).
+
+from odoo.tests import common
+from odoo.exceptions import ValidationError
+
+
+class EventSession(common.SavepointCase):
+
+    @classmethod
```
NL66278, line 13: Also remove the seats stuff from the tests.

pedrobaeza:
Sorry, but you can't rip off all that features from the module. They are needed the same as you have the `event` module as only one module, not `event` and `event_registration`, so :-1: to this migration while it doesn't preserve all the features we added in v10.

NL66278:
@pedrobaeza For us, and probably others, the idea of sessions is that you have one event where you register for (for instance a python course), but that is spread out over sessions (let us say for seven weeks). You do not register for each separate lesson.

But I do understand what you are saying. The usecase of the original authors of the module would be probably be to use an event as a kind of template, to generate several sessions with separate registration etc. Although I do not quite see why you would not generate events then and not sessions.

Would it be a solution to do the migration by renaming the module migrated/changed by @ntsirintanis into event_session_base, and then have event_session extend this to add the concepts of separate registration/availability etc.?

pedrobaeza:
@NL66278 we are the original authors of the module, so I have very clear what is the goal of the module ;) It's not that, it's for having the concept of session, which is another pass of the same event. For example, of theater shows, cinema, and so on, so registration is perfectly fit for doing that at session level.

What you are saying is that you pretend to use this as tracks, not sessions, so this is not intended for that.

And we designed this to have the opportunity to register to an event or to a session, so these features don't block you usage, and so you shouldn't split out this.

NL66278:
@pedrobaeza I agree then that we will have to provide the full migration. For now consider it as a work in progress. The underlying code for registrations changed a lot, so this will take more time. FYI we looked at tracks but they did not fit our usecase. Tracks are more things happening in parallel, like all the different tracks at the Odoo days.

pedrobaeza:
OK, thanks for reconsidering it. I have used in the past tracks for both things: parallel and linear "sessions", but of course it depends on the rest of requirements you have for that sessions. If you share them, maybe we can do a brain-storming about this.

ntsirintanis:
I have gradually reintroduced all features that were removed for the initial PR. Please take a look @NL66278

alexschubert (changes requested):
Thanks for your contribution!

During functional testing, the following error appeared when trying to add a session to a newly created event:

> Uncaught Error: KeyError: 'id'
> http://3349535-127-b590c5.runbot3.odoo-community.org/web/content/418-c35d703/web.assets_backend.js:369
> Traceback:
> Error: KeyError: 'id'
>     at F.__getitem__ (http://3349535-127-b590c5.runbot3.odoo-community.org/web/content/418-c35d703/web.assets_backend.js:369:161)
>     at F.__getattr__ (http://3349535-127-b590c5.runbot3.odoo-community.org/web/content/418-c35d703/web.assets_backend.js:370:69)
>     at F.__getattribute__ (http://3349535-127-b590c5.runbot3.odoo-community.org/web/content/418-c35d703/web.assets_backend.js:203:38)
>     at Object.py.PY_getAttr (http://3349535-127-b590c5.runbot3.odoo-community.org/web/content/418-c35d703/web.assets_backend.js:167:176)
>     at Object.py.evaluate (http://3349535-127-b590c5.runbot3.odoo-community.org/web/content/418-c35d703/web.assets_backend.js:270:11)
>     at Object.py.evaluate (http://3349535-127-b590c5.runbot3.odoo-community.org/web/content/418-c35d703/web.assets_backend.js:265:111)
>     at Object.py.evaluate (http://3349535-127-b590c5.runbot3.odoo-community.org/web/content/418-c35d703/web.assets_backend.js:267:99)
>     at Object.py.eval (http://3349535-127-b590c5.runbot3.odoo-community.org/web/content/418-c35d703/web.assets_backend.js:272:284)
>     at http://3349535-127-b590c5.runbot3.odoo-community.org/web/content/418-c35d703/web.assets_backend.js:380:136
>     at Function._.each._.forEach (http://3349535-127-b590c5.runbot3.odoo-community.org/web/content/412-150dd49/web.assets_common.js:12:558)

Steps to reproduce:
1. Open the "Events" module
2. Click on the "Create" action
3. Without editing the event any further, add a session to the "Email Schedule" notebook page
4. Click on the Session dropdown
5. The error appears


Apart from this error, I was able to validate the general functionality of the module in addition to features like separate session registration and seat availability per session.

ntsirintanis:
@alexschubert Thanks a lot for your review and the time spent testing the module. I have corrected the error by hiding event_mail_ids tree if there are no sessions associated to the event. This is different than 10.0, where one could actually create a session without having an event record. The way it is implemented now is more logical and clear in my opinion.

event_session/views/event_session_view.xml
```diff
@@ -0,0 +1,148 @@
...
+            </calendar>
+        </field>
+    </record>
+
+    <!-- EVENT.SESSIONS ACTIONS -->
+    <record id="act_event_session_form" model="ir.actions.act_window">
+        <field name="res_model">event.session</field>
+        <field name="view_type">form</field>
+        <field name="name">Sessions</field>
+        <field name="view_mode">tree,form,calendar</field>
```
chienandalu: It'd be hand to have a pivot as well

chienandalu (commented):
Looks good (code and runbot). Just a minor comment:

ntsirintanis:
@chienandalu Thanks for your review and suggestion. Event_session has an option for pivot view now.

event_session/models/event.py
chienandalu: @ntsirintanis What's the reason behind introducing this in the event object?

ntsirintanis: @chienandalu Ah, that's  remnant from some experimentation I did with the module. Thanks for spotting this, I have pushed a new commit that removes this overwrite.

NL66278 (approved):
Tested in practice and looks good to me.

chienandalu (approved):

PR #128, opened 2018-12-10
[IMP] partner_event: support multi-company

chienandalu:
- Add support for multi-company, so we create or associate the partners to the company configured in the event.

cc @Tecnativa

### DIFF
--- a/partner_event/models/event_registration.py
+++ b/partner_event/models/event_registration.py
@@ -25,6 +25,7 @@ class EventRegistration(models.Model):
             'name': vals.get('name') or vals.get('email'),
             'email': vals.get('email', False),
             'phone': vals.get('phone', False),
+            'company_id': vals.get('company_id', False),
         }
 
     @api.model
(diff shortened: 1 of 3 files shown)


partner_event/models/event_registration.py
```diff
@@ -34,16 +35,19 @@ def create(self, vals):
             Event = self.env['event.event']
             # Look for a partner with that email
             email = vals.get('email').replace('%', '').replace('_', '\\_')
+            event = Event.browse(vals['event_id'])
+            company_id = event.company_id and event.company_id.id
```
carlosdauden:
```suggestion
            company_id = event.company_id.id
```

partner_event/models/event_registration.py
```diff
@@ -34,16 +35,19 @@ def create(self, vals):
...
             attendee_partner = Partner.search([
-                ('email', '=ilike', email)
+                ('email', '=ilike', email),
+                ('company_id', '=', company_id),
```
carlosdauden, line 43:
```suggestion
                '|',
                ('company_id', '=', company_id),
                ('company_id', '=', False),
```

carlosdauden (changes requested):

partner_event/models/event_registration.py
carlosdauden: Other better solution can be a second search with ('company_id', '=', False) after first search if it don't find results

chienandalu: Well, indeed if the company_id is False, that means that the partner is shared across companies, so it should be fine, isn't it?

carlosdauden (approved):

partner_event/models/event_registration.py
```diff
@@ -34,16 +35,21 @@ def create(self, vals):
...
-            event = Event.browse(vals['event_id'])
             if attendee_partner:
                 vals['name'] = vals.setdefault('name', attendee_partner.name)
                 vals['phone'] = vals.setdefault(
                     'phone', attendee_partner.phone)
             elif event.create_partner:
                 # Create partner
+                vals['company_id'] = company_id
                 attendee_partner = Partner.sudo().create(
                     self._prepare_partner(vals))
```
pedrobaeza, line 54: This is unrelated, but isn't this going to generate warning due to invalid fields that are on `event.registration`, but not on `res.partner`?

pedrobaeza (approved):

partner_event/models/event_registration.py
pedrobaeza: Yes, but it's for the improbable case of having 2 equal partners, one with the company set, and another without, giving priority to the specific company partner, but I think it's to fine grain a lot. What do you think?

rafaelbn (approved):
Tested functionally in runbot :+1:

PR #129, opened 2019-02-11
Migrate script sends cr and version, not env

Martronic-SA:


pedrobaeza:
Not really, openupgradelib passes `env` by default in v10+:

https://github.com/OCA/openupgradelib/blob/926b3dcfaeebe73c0c6e2d29d157760b86889912/openupgradelib/openupgrade.py#L1395

Martronic-SA:
Thank you for your reply. I've had an error on env.cr.
So probably my openupgralib version is not upto date.

Martronic-SA:
the "use_env=True" option is obsolete then?

pedrobaeza:
Yes for 10+

PR #130, opened 2019-02-14
[12.0][MIG] website_event_share

cristinamartinrod:
cc @Tecnativa


pedrobaeza (commented):
Use new readme by fragments system

pedrobaeza (commented):
Trying on runbot, I see the icons and they are working properly. The only doubt I have is about the Graph system I have read about in v12, that allows to select which image to use. How to use it here?

cristinamartinrod:
@pedrobaeza That option set in the "Promote > Optimize SEO" section on top menu.

pedrobaeza (approved):
OK, I see, as that should be an option that we set fixed for all the shares. Great, thanks.

website_event_share/readme/USAGE.rst
```diff
@@ -0,0 +1,7 @@
+1. Go to the website events menu.
+2. Open one of your events.
+3. Click on the customize button.
+4. Toggle the share button option to enable/disable the share links block.
```
yajo: In general it's better to use `#.` for bullet lists, to use automatic numeration.

yajo (approved):
It looks great!

captura de pantalla de 2019-02-15 08-40-00

Maybe you want to display that image in the README. :point_up:

PR #133, opened 2019-03-13
[ADD] website_event_questions_free_text: Free Text Answers on Events Questions

pedrobaeza:
This module allows to define free text answers for the event registrations in the website, so you can answer freely without the need of selecting a closed list of choices.

Configuration
=============

1. Go to *Events > Events*
2. Create or select one event.
3. On the "Questions" page, create a new question or edit an existing one.
4. Mark the check "Free Text" for marking that question for having a free text answer.

Usage
=====

1. Go to the website.
2. Go to "Events" section and select one of them.
3. Select ticket quantity and click on "Register".
4. On the popup dialog, fill the text area for the free text question.

Known issues / Roadmap
======================

* Add tests (tour) to the module.

cc @Tecnativa


website_event_questions_free_text/controllers/main.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def _process_registration_details(self, details):
+        """Add free answers to the registration"""
+        registrations = super()._process_registration_details(details)
+        for registration in registrations:
+            free_answers = []
+            for key, value in registration.items():
+                if key.startswith('answer_free_text-'):
+                    free_answers.append((0, 0, {
+                        'question_id': int(key[17:]),
```
victor-ditay, line 17: Just for my understanding as I've already tested on Runbot and works as it's expected.
This means that 'answer_free_text-' should be always like 'answer_free_text-[integer]' ?

victor-ditay (approved):

website_event_questions_free_text/controllers/main.py
pedrobaeza: Yes, that's what post method passes due to the QWeb template is put that way.

Tardo (approved):

PR #134, opened 2019-03-14
[ADD] website_event_questions_by_ticket: Conditional Events Questions

pedrobaeza:
This module allows to set that an event question is only asked for certain tickets.

If the question is general, then having one of the allowed tickets will serve for showing it. If the question is specific to each attendee, then the evaluation will be done for each of the tickets.

Configuration
=============

1. Go to *Events > Events*
2. Create or select one event.
3. On the "Questions" page, create a new question or edit an existing one.
4. Fill the field "Limited to tickets" for restricting the question to that ticket. Leave it blank for making it available for all tickets.

Usage
=====

1. Go to the website.
2. Go to "Events" section and select one of them.
3. Select ticket quantity and click on "Register".
4. On the popup dialog, the questions linked to that ticket and those without restrictions will be shown.

Known issues / Roadmap
======================

* Add tests (tour) to the module.

@Tecnativa


victor-ditay (approved):
LGTM tested on Runbot

Tardo (approved):

Tardo:
Perhaps can be good add an option to allow empty initialization of list responses... ?¿?¿?

pedrobaeza:
@Tardo I think that would be another different module, as this is only allowing to set questions depending on the ticket. Merging.

PR #135, opened 2019-03-15
[10.0] website_event_require_login: New module

chienandalu:
Require login for event registration from the website.

cc @Tecnativa


pedrobaeza:
And the event is going to be visible although not registrable?

chienandalu:
Yes, the main object is to have a registered partner in the registration.

pedrobaeza:
OK, but add to the roadmap the feature I'm talking about

website_event_require_login/controllers/main.py
```diff
@@ -0,0 +1,18 @@
...
+# Copyright 2019 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import http
+from odoo.addons.website_event.controllers.main import WebsiteEventController
+
+
+class RequireLoginToRegister(WebsiteEventController):
+
+    @http.route()
```
yajo: Have you tested just overriding this and putting `@http.route(auth="user")`? If I don't remember bad, doing that autoredirects to `/web/login` for unauthenticated users. :thinking:

website_event_require_login/controllers/main.py
```diff
@@ -0,0 +1,18 @@
...
+    def registration_new(self, event, **post):
+        if event.website_require_login:
```
yajo: Is there some reason why you aren't actually checking if the user is logged in or not?

website_event_require_login/controllers/main.py
chienandalu: Sure, but we don't want to restrict it to all users

website_event_require_login/controllers/main.py
chienandalu: Sure :man_facepalming:

chienandalu:
@Yajo Changes done

website_event_require_login/controllers/main.py
```diff
@@ -0,0 +1,19 @@
...
+
+from odoo.http import request, route
+from odoo.addons.website_event.controllers.main import WebsiteEventController
+
+
+class RequireLoginToRegister(WebsiteEventController):
+
+    @route()
+    def registration_new(self, event, **post):
+        public_user = request.env.user == request.env.ref('base.public_user')
```
yajo: The public user should be extracted from `request.website.user_id` instead.

website_event_require_login/controllers/main.py
```diff
@@ -0,0 +1,19 @@
...
+        if public_user and event.website_require_login:
+            return request.env['ir.ui.view'].render_template(
```
yajo, line 15: IMHO it would be better to just redirect the user to `"/web/login?redirect=%s" % event_url`. Don't you think so? We could save the need to maintain a separate template, and we'd save a click for the user.

yajo (changes requested):

website_event_require_login/controllers/main.py
chienandalu: Sure the redirect is needed to keep the user on track, but the popup is necessary for the user to know that the event requires login. To kick him off without warning would be misleading I thing :)

website_event_require_login/README.rst
```diff
@@ -0,0 +1,96 @@
...
+
+Authors
+~~~~~~~
+
+* Tecnativa
+
+Contributors
+~~~~~~~~~~~~
+
+* David Vidal <david.vidal@tecnativa.com>
```
rafaelbn: I have spent some hours with this matter... if you consider I'm contribur please add me. Thanks

rafaelbn (approved):
Tested functionaly :+1: 

Please @Yajo review again it work fine!

yajo (approved):

emagdalenaC2i:
_Known issues / Roadmap
An interesting option could be to be able to hide events to unregistered users._
@chienandalu and @pedrobaeza in v11 and v12 you already have that option
image

PR #138, opened 2019-03-21
12.0 mig event registration multi qty

chienandalu:
This module allows to make registration to events with more than one attendee.

This was the unique behavior on versions 8 and 9. As it's useful for some use
cases, this module allows both ways of registering attendees.

cc @Tecnativa


pedrobaeza (approved):
Should we include in the mailing template the number of attendees?

event_registration_multi_qty/README.rst
```diff
@@ -0,0 +1,98 @@
...
+   :local:
+
+Configuration
+=============
+
+To configure the possibility of registering mutiple attendees:
+
+#. Go to *Events > Events*.
+#. Create one event.
+#. Go to *Registration tab* and check "Registration Multi Qty" to activate
```
victor-ditay, line 45: LGTM but seems this need update. I tested on Runbot and the option is now the default behavior. 

image

victor-ditay (approved):

PR #136, opened 2019-03-20
[12.0] event_session: Migration to 12.0

chienandalu:
This module allows to create sessions associated with events.

Since the dependency to event_mail was embedded in this module, we've migrated it as well as the core features of 12.0 doesn't fully cover all the cases.

- [x] event_mail https://github.com/OCA/event/pull/145

cc @Tecnativa

PR #140, opened 2019-03-21
 [12.0] event_session_registration_multi_qty: Migration to 12.0

chienandalu:
Depends on:

- [x] https://github.com/OCA/event/pull/136 `event_session`
- [x] https://github.com/OCA/event/pull/138 `event_registration_multy_qty`

This module allows to make registration to sessions with multiple attendees.

cc @Tecnativa


event_session_registration_multi_qty/models/event_session.py
```diff
@@ -1,20 +1,23 @@
...
+# Copyright 2017 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl-3.0).
+from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
 
 
 class EventSession(models.Model):
     _inherit = 'event.session'
 
+    registration_multi_qty = fields.Boolean(
```
pedrobaeza: Why adding this related?

chienandalu: Sure, it's useless. There was a view that had a reference to this field, although it wasn't declared, so...

pedrobaeza:
Please rebase

chienandalu:
This would be ready to merge

pedrobaeza (approved):

chienandalu:
ping @rafaelbn

victor-ditay (approved):
LGTM but no way to test in runbot

event_session_registration_multi_qty/__init__.py
```diff
@@ -0,0 +1,2 @@
+from . import models
+from . import tests
```
emagdalenaC2i: I think you don't need to import tests

event_session_registration_multi_qty/tests/__init__.py
```diff
@@ -0,0 +1 @@
+from . import test_event_session_registration_multi_qty
```
emagdalenaC2i, line 1: You can use a smaller filename instead of test_event_session_registration_multi_qty

event_session_registration_multi_qty/tests/test_event_session_registration_multi_qty.py
```diff
@@ -0,0 +1,93 @@
+# Copyright 2017-19 Tecnativa - David Vidal
```
emagdalenaC2i, line 1: You can use a smaller filename instead of test_event_session_registration_multi_qty

emagdalenaC2i (approved):
Code review and tested in runbot

event_session_registration_multi_qty/__init__.py
pedrobaeza: @chienandalu this is true.

event_session_registration_multi_qty/tests/test_event_session_registration_multi_qty.py
pedrobaeza: Well, the name is test_ + module name, and there's no problem in that scheme IMO.

emagdalenaC2i: I know, but the module name is too long and there are other modules that use a shorter name in this test files
Like https://github.com/OCA/account-payment/blob/12.0/account_payment_return_import_iso20022/tests/test_import_iso20022.py

chienandalu: Sorry, I don't see a problem with it, and it's a convention to us. :)

event_session_registration_multi_qty/__init__.py
chienandalu: Sure :)

chienandalu:
Changes done

PR #143, opened 2019-04-18
[MIG] event_email_reminder: Migrate to v12

yajo:
Standard migration.

@Tecnativa


pedrobaeza:
CIs red

yajo:
Let's see now.

pedrobaeza:
Still no luck

yajo:
Oops, let's see now.

yajo:
This time I think is the good one. :stuck_out_tongue:

event_email_reminder/models/event.py
```diff
@@ -42,27 +41,26 @@ def run_event_email_reminder(
...
+        limit_date = today + timedelta(days=days)
         if draft_events:
             domain = [('state', 'in', ['draft', 'confirm'])]
         else:
             domain = [('state', '=', 'confirm')]
         if not near_events:
             domain.extend([
-                ('date_begin', '>=', '%s 00:00:00' % limit_date),
-                ('date_begin', '<=', '%s 23:59:59' % limit_date),
+                ('date_begin', '>=', datetime.combine(limit_date, time_start)),
```
pedrobaeza: AFAIK, domains still use string format, so I think this is working by chance as the `repr` of the datetime matches the format, but that can be assured. You should do `fields.Datetime.to_string(...`

pedrobaeza (changes requested):
Please fix the cron data XML:
```
2019-04-23 11:52:09,263 160 WARNING openerp_test odoo.models: ir.cron.create() with unknown fields: args, function
```

event_email_reminder/models/event.py
yajo: I was curious. Actually it's the other way around: [if you use a string, it's converted to a datetime object. If you use a datetime, it's used raw](https://github.com/odoo/odoo/blob/9617e79e583aeb3d6d983045c5934dae28dce0c1/odoo/osv/expression.py#L1101-L1115).

It's interesting, though, that I could have saved the ~~`00:00:00` and `23:59:59`~~ `datetime.combine()` thingies. :thinking:

yajo:
Done.

event_email_reminder/models/event.py
pedrobaeza: Well, seeing the code, it seems both are equally accepted, so good for that "polymorphism".

pedrobaeza:
Now runbot and Travis are red.

yajo:
Yikes, a typo.

pedrobaeza (approved):

victor-ditay (approved):

yajo:
Is it ugly if I merge it myself? :thinking:

pedrobaeza:
I save you the pain, hehe

PR #145, opened 2019-05-06
[12.0] event_mail: Migration to 12.0

chienandalu:
This module allows to define mail settings for events. By default the emails
scheduler has been deactivated.

You can create mail scheduler templates for events and select one by default
in event settings.

cc @Tecnativa


Tardo (approved):

Tardo:
I don't test if the mail is really sent...

chienandalu:
> I don't test if the mail is really sent...

Well, this module has no influence on that. It only assigns templates to an event

yajo (approved):

event_mail/hooks.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2019 David Vidal <david.vidal@tecnativa.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+
+def pre_init_hook(cr):
+    """Create company event_mail_template_id column in advance"""
```
pedrobaeza: Why are you doing this? Don't document the "what", which can be seen reading the code, but the "why"

event_mail/models/event.py
```diff
@@ -10,8 +8,7 @@ class EventEvent(models.Model):
 
     @api.model
     def _default_event_mail_template_id(self):
-        return self.env['ir.values'].get_default(
-            'event.config.settings', 'event_mail_template_id')
+        return self.env.user.company_id.event_mail_template_id
```
pedrobaeza: You should use the default `company_id` value (that can be different than the user one).

event_mail/models/event.py
```diff
@@ -33,14 +30,7 @@ def _onchange_event_mail_template_id(self):
...
-        event_mail_template_id = self.env['ir.values'].get_default(
-            'event.config.settings', 'event_mail_template_id')
-        if event_mail_template_id:
-            return super(EventEvent, self)._default_event_mail_ids()
-        else:
-            return []
+    @api.onchange('event_type_id')
+    def _onchange_type(self):
+        if self.event_type_id and self.event_mail_template_id:
+            self._onchange_event_mail_template_id()
```
pedrobaeza: Why should we reset templates on event type change?

event_mail/models/event_type.py
```diff
@@ -0,0 +1,23 @@
...
+
+class EventType(models.Model):
+    _inherit = 'event.type'
+
+    @api.model
+    def _get_default_event_type_mail_ids(self):
+        if self.env.context.get('by_pass_config_template', False):
+            return super()._get_default_event_type_mail_ids()
+        event_mail_template_id = (
+            self.env.user.company_id.event_mail_template_id)
```
pedrobaeza: Use default `company_id`

pedrobaeza (changes requested):

event_mail/models/event.py
yajo: I don't understand this. Why would user from company A want a default for company B? :thinking:

pedrobaeza: We don't know, but let the ORM do their job. If field `company_id` is set by default to any other value, the same should be used here. Any way, don't think only on different companies. Think on parent + child companies for example.

yajo: But are you talking about the `company_id` default value in `event.event`, or in `res.users`?

pedrobaeza: In `event.event`. Quick code:
```python
return self._default_get('company_id').event_mail_template_id
```

yajo: I understood you were talking about the default company for the database, not for the event. You're right, thanks! :blush:

event_mail/hooks.py
chienandalu: Right. It's because it's being called in a field default value, so it's needed in advance

event_mail/models/event.py
chienandalu: This was bad implemented indeed. It's an onchange from the event module. New implementation coming so we just override it if a scheduling template is set

event_mail/models/event.py
chienandalu: Ok, changed :)

event_mail/models/event_type.py
chienandalu: changed

event_mail/hooks.py
pedrobaeza: Do you mean for not having any default value in it?

chienandalu: No, because the column doesn't exist yet in the model res.company.

pedrobaeza: Uhm, curious. I would say that this shouldn't happen, but it seems Python order sometimes matters... Maybe you can change the import order for avoiding this hook.

chienandalu: That was it :)

pedrobaeza (approved):

PR #152, opened 2019-06-03
[MIG] website_event_questions_template: Migration to 12.0

victor-ditay:
cc @Tecnativa


victor-ditay:
@sergio-teruel Please have a check

pedrobaeza (changes requested):
We used the trick of loading the questions in v10 with a button because there were no proper support for onchanges in nested one2many, but now I think the support is correct. Can you please switch back to fill things in onchange for not needing to save + press button?

pedrobaeza:
Other minor thing is that a field with a number that I suppose is the sequence is shown when defining the question template.

victor-ditay:
I will refactor based on your comments @pedrobaeza

victor-ditay:
@pedrobaeza I pushed a new WIP commit so you could test if could work onchage option

victor-ditay:
>Other minor thing is that a field with a number that I suppose is the sequence is shown when defining the question template.

@pedrobaeza I removed the previous commit, here is the reason https://github.com/odoo/odoo/issues/20163#issuecomment-505055793

website_event_questions_template/models/event_question_template.py
```diff
@@ -1,5 +1,4 @@
-# -*- coding: utf-8 -*-
-# Copyright 2017 Tecnativa - Sergio Teruel
+# Copyright 2017-2019 Tecnativa - Sergio Teruel
```
pedrobaeza: Sergio hasn't done anything in 2019, so you shouldn't change its copyright. In any case, yours should be added in 2019, but any way, this file hasn't been modified, so you don't have to change whole copyright. This applies to all the files incorrectly modified.

website_event_questions_template/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,5 @@
+* Tecnativa (https://www.tecnativa.com)
+
+    * Sergio Teruel <sergio.teruel@tecnativa.com>
+
```
pedrobaeza: Add you as contributor

pedrobaeza (changes requested):

victor-ditay:
Changes done @pedrobaeza

website_event_questions_template/views/event_view.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<!-- Copyright 2017 Tecnativa - Sergio Teruel
+<!-- Copyright 2017-2019 Tecnativa - Sergio Teruel
```
pedrobaeza: This still remains

website_event_questions_template/views/website_event_questions_template_view.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<!-- Copyright 2017 Tecnativa - Sergio Teruel
+<!-- Copyright 2017-2019 Tecnativa - Sergio Teruel
```
pedrobaeza: This still remains

victor-ditay:
Changes done @pedrobaeza

pedrobaeza (approved):

sergio-teruel (approved):

PR #153, opened 2019-06-04
[12.0][MIG] event_contact

Tardo:
cc @Tecnativa


event_contact/tests/test_event.py
```diff
@@ -1,35 +1,55 @@
...
+        })
+        self.type2 = self.env['event.type'].create({
+            'name': 'Event Type Test 2',
+            'contact_ids': [(6, False, [self.partner3.id, self.partner4.id])],
+        })
+
+        self.event1 = self.env['event.event'].create({
+            'name': 'Event Test 1',
+            'date_begin': datetime.now(),
+            'date_end': datetime.now() + timedelta(days=3),
```
chienandalu: For this tests I think you can use a fixed date as it's not really relevant to the overcome

event_contact/tests/test_event.py
```diff
@@ -1,35 +1,55 @@
...
+        })
 
     def test_event_onchange_type_contacts_empty(self):
         """You get default contacts from type."""
-        self.event1.event_type_id = self.type2
+        self.event1.event_type_id = self.type2.id
```
chienandalu: Why do need to assign like this?

event_contact/views/event_type_view.xml
```diff
@@ -9,8 +10,16 @@
     <field name="model">event.type</field>
     <field name="inherit_id" ref="event.view_event_type_form"/>
     <field name="arch" type="xml">
-        <xpath expr="//group">
-            <field name="contact_ids" widget="many2many_tags"/>
+        <xpath expr="//div[@name='event_type_attendees']" position="inside">
+            <div class="col-12 col-lg-6 o_setting_box" name="event_type_contacts">
+                <!--div class="o_setting_left_pane">
+                    <field name="contact_ids" widget="many2many_tags"/>
+                </div-->
```
chienandalu: Remove commented code :)

chienandalu (changes requested):
Thanks, @Tardo I tested on runbot and it works fine. Just some comments:

event_contact/tests/test_event.py
Tardo: In Odoo the two forms are used interchangeably... i like use 'id' because internally uses this value to set the field.

Tardo:
@chienandalu thx, changes done!

chienandalu (approved):
Great! :)

event_contact/tests/test_event.py
pedrobaeza: Yes, and you avoid some problems like different month length, leap years...

pedrobaeza (approved):

PR #155, opened 2019-06-28
[MIG] website_event_questions_by_ticket: Migration to 12.0

dioalves:
The code is the same as the version 11.0, only changed the version on manifest file.


pedrobaeza:
Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0

ioans73 (approved):

PR #157, opened 2019-07-05
[FIX] vals of event registration

Montxu:
attendee_partner_id is not avaliable in v12.

### DIFF
--- a/partner_event/models/event_registration.py
+++ b/partner_event/models/event_registration.py
@@ -46,7 +46,7 @@ class EventRegistration(models.Model):
                 # Create partner
                 attendee_partner = Partner.sudo().create(
                     self._prepare_partner(vals))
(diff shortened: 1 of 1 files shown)


pedrobaeza:
It's defined in this module... https://github.com/OCA/event/pull/157/files#diff-22bed8d3e426204a537b0d8e18f9f8e3R16

Montxu:
I'm sorry it's my mistake, it's true.
In this way the registration of participants, even if I find the contact in my system and if it is not logged in which partner_id will always be public user, isn't it?

pedrobaeza:
Yes, because Odoo now distinguishes between the partner who does the registration and the attendee itself

pedrobaeza:
Closing as incorrect.

Montxu:
Thanks for explanation and your time.

pedrobaeza:
You are welcome.

PR #158, opened 2019-08-22
[FIX] event_session: conflicting configuration

chienandalu:
- The config is provided by event_mail and the model mismatches the original one.

cc @Tecnativa


pedrobaeza (approved):
Shouldn't a migration script that removes the ir.default should be provided?

chienandalu:
Well, this shouldn't ever have been here. Is a migration blunder introduced in v11.

pedrobaeza:
OK, let's fast track it

/ocabot merge patch

PR #160, opened 2019-08-26
[MIG] crm_lead_to_event_registration: Migration to v12

victor-ditay:


victor-ditay:
No need migrate 
cc @Tecnativa

PR #161, opened 2019-08-26
[MIG] crm_lead_to_event_registration: Migration to v12

victor-ditay:
cc @Tecnativa


pedrobaeza (approved):

rafaelbn (commented):
In the wizard we should filter for upcoming events, not old ones

pedrobaeza:
@rafaelbn you mustn't restrict that selection in any way, as you don't know the intention of that dump.

rafaelbn:
The attendee should be the contact and not the company

pedrobaeza:
That's not something of this module, but maybe `partner_event`.

rafaelbn:
> @rafaelbn you mustn't restrict that selection in any way, as you don't know the intention of that dump.

@pedrobaeza it is an opportunity so the excepted by default is new one. The exception is register to old ones, so if you want to register an exception in an old event you should:

- Search more
- delete filter upcoming
- look for the old event

pedrobaeza:
That's the problem, @rafaelbn, if you restrict here to upcoming, you won't have any way to select them.

victor-ditay:
@rafaelbn Please review

rafaelbn (approved):

PR #162, opened 2019-08-26
[MIG] event_registration_partner_unique: Migration to v12

victor-ditay:


pedrobaeza (approved):
You can squash bot commits (for README and icon) with nearest previous migration commit

victor-ditay:
Changes done @pedrobaeza

rafaelbn (commented):
- Register any contact to an event as attendee
- Go to the website and register the same contact

You get 500: Internal Server Error

rafaelbn:
@victormmtorres second issue I found is that from Sales -> Quotation I have duplicated an attendee ( cc @chienandalu )



It shouldn't be able.

This is because module must check not when confirming the sale order




Check after confirming in the wizard for the attendees

victor-ditay:
> Register any contact to an event as attendee
Go to the website and register the same contact
You get 500: Internal Server Error

@rafaelbn @chienandalu This happens on v11 as well

About second issue, I couldn't reproduce, perhaps we could check tomorrow. I have doubts is the same partner as you show in your screenshot https://user-images.githubusercontent.com/8736623/63792616-35b7d800-c8fe-11e9-850a-2bb48f9fc79c.png

Attendee Partner default_get is not the same:
first is: Deco Addict, Floyd Steward
second is: Menganito

chienandalu (commented):
> second issue I found is that from Sales -> Quotation I have duplicated an attendee

I'd say that's the right behavior, as the checked field is the attendee_partner_id, no the partner_id, who should be able to register multiple attendees as in the case you described

chienandalu (commented):
> Register any contact to an event as attendee
> Go to the website and register the same contact
> You get 500: Internal Server Error

Seem like the exception is raising below the surface. Maybe we could get rid of the custom exceptions defined in this module for starters. (@Yajo do you remember why did you implement them?)

yajo:
Usually if I do custom exceptions is to differentiate them in tests (with assertRaises).

pedrobaeza:
It's better to not use custom exceptions, but standard ones and only differentiate them by message.

rafaelbn:
Closing an reopening because runbot says Internal Server Error and I cannot rebuild

rafaelbn:
Please @victormmtorres update README with this "issues" not related in the ROADMAP and let go ahead, thanks! Please add me as contributor, I should be before

victor-ditay:
>Please add me as contributor, I should be before

@rafaelbn You already there https://github.com/OCA/event/pull/162/files#diff-c9fb6277c1657c54d6a076d5395a747eR67

victor-ditay:
Please review @rafaelbn

rafaelbn (approved):

pedrobaeza:
@rafaelbn please review together with @victormmtorres the README, as it contains some gramatical errors and it's not very understandable IMO.

rafaelbn (dismissed):
Improve README

victor-ditay:
Changes done @rafaelbn @pedrobaeza

PR #164, opened 2019-08-30
Use logged-in users name and email in event registration automatically

chrisandrewmann:
Update the website_event_registration_attendee_details view to use the logged in user's name and email address instead of allowing freetyped input fields to prevent human error.

Combined with requiring a login for an event, helps simplify registering process as users shouldn't need to re-enter their details. Phone is hidden also as not needed if they are linked to a partner record.

### DIFF
--- a/website_event_require_login/__manifest__.py
+++ b/website_event_require_login/__manifest__.py
@@ -13,7 +13,7 @@
     'data': [
         'views/event_views.xml',
         'views/website_event_templates.xml',
-		'views/website_event_registration_attendee_details_views.xml'
+        'views/website_event_registration_attendee_details_views.xml'
     ],
     'installable': True,
     'license': 'AGPL-3',
--- a/website_event_require_login/__manifest__.py
+++ b/website_event_require_login/__manifest__.py
@@ -2,7 +2,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 {
     'name': 'Website Event Require Login',
-    'version': '12.0.1.0.0',
+    'version': '12.0.1.1.0',
     'author': 'Tecnativa, '
               'Odoo Community Association (OCA)',
     'website': 'https://github.com/OCA/event',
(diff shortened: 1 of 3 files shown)


rafaelbn:
Hi @chrisandrewmann , Thanks for this PR but this functionality in not desired in this module. So you have to make a new module  website_event_attendee_with_login o similar.

Because this functionality wants that anonymous users cannot register, just that. Because attendees shouldn't be the same of the user which register them.

My best

chrisandrewmann:
Hi there,
I'd disagree with the comment made. I think this is much needed functionality that fixes the issue of inconsistency when registering for an event that requires login. You should only be able to register for the event under your OWN name and email address, not on behalf of others if it is implied that only approved users/partners can register.
Propose a compromise might be a custom field added to the attendee registration module and view to allow "notes" about their particular registration request.

pedrobaeza:
@chrisandrewmann I agree with @rafaelbn that the feature you are adding, although interesting, must be at least apart, due to several things:

* On standard behavior, you are able to register other attendees names within your registration. There are then 2 roles: attendees and the people who register them. This module forces to login for not having an anonymous second role, but the first one should continue the same.
* If you register more than 1 place, should all go to the same person? That's limiting the info you will receive. You are usually interested of having all the attendee names.
* This behavior should be opted-in at event level, for having the granularity of choosing which events doesn't allow to register other person.

With all of this, you can create a new module with the feature, having into account my remarks, but not inside this existing module.

rafaelbn (changes requested):
> I think this is much needed functionality

I don't say that is a functionality not needed, I say that you cannot add that functionality in this module because several reasons. Odoo is modular we should maintain like that.

> that fixes the issue of inconsistency when registering for an event that requires login

This is your opinion. There is not inconsistency and nothing to fix, it is not a bug. You want to add functionality not fix existing one.

> You should only be able to register for the event under your OWN name and email address, not on behalf of others if it is implied that only approved users/partners can register.

This is again you opinion. I have registered to my employees to the OCA Code Sprint and to the Odoo Experience because company pay. But there are lots or more reasons.

As commented you should create a new module that can depend on this one

PR #165, opened 2019-09-01
Add new module which allows restricting registering of event tickets to logged-in users

chrisandrewmann:
Add new module which allows restricting registering of event tickets from website to ONLY the logged in user (email and name).

This forces consistency to ensure that all tickets MUST be registered from a logged-in user.
For companies using an approval scheme to only allow approved users, this is important to prevent unknown members of the general public registering for events.
For example when the event is a webinar for partners only.


chrisandrewmann:
Hi there. Sorry I've added that email to my github account now and submitted an agreement on Friday.

PR #166, opened 2019-09-01
[ADD] new module website_event_attendee_with_login

chrisandrewmann:
Add new module website_event_attendee_with_login to require attendee registrants to have their own accounts.

All tickets will be under the name/email of the logged-in user preventing them from signing up members of the public.
This is important for companies who only allow approved members to attend events such as webinars.


emagdalenaC2i:
There are the module website_event_require_login for Odoo 11 "https://github.com/OCA/event/tree/11.0/website_event_require_login"
You should migrate it instead of create a new one.

chrisandrewmann:
Hello there,
This is not the same module, it offers different functionality on the attendee registering for a ticket, preventing them from specifying different email/name to their logged-in user account.
website_event_require_login gives different functionality and you can speak to the owners of that module who rejected my suggestion to add this functionality to their module.
Regards,
Chris

chrisandrewmann:
Please see this previous PR which explains all:
[https://github.com/OCA/event/pull/164](https://github.com/OCA/event/pull/164)

website_event_attendee_with_login/.gitignore
```diff
@@ -0,0 +1,352 @@
+## Ignore Visual Studio temporary files, build results, and
```
chienandalu, line 1: Use repo .gitignore

chienandalu (commented):
Why not simply bypass the ticket form fill and get the logged in user for it? You could depend on website_event_require_login to do so. I'd rename the module to something like website_event_force_attendee_from_login as well.

PR #170, opened 2019-10-08
[IMP] partner_event: add data to old registrations

chienandalu:
When installing the module, old registrations attendee partner should be set in the same way the new ones do.

cc @Tecnativa

### DIFF
--- a/partner_event/hooks.py
+++ b/partner_event/hooks.py
@@ -1,0 +1,22 @@
+# Copyright 2019 David Vidal
+# License AGPL-3 - See https://www.gnu.org/licenses/agpl-3.0.html
+from odoo import api, SUPERUSER_ID
+
+
+def post_init_hook(cr, registry):
+    """Preload proper attendee partner for existing registrations using
+       the same rules the module does"""
+    with api.Environment.manage():
+        env = api.Environment(cr, SUPERUSER_ID, {})
+        attendees_emails = env['event.registration'].read_group([
+            ('email', '!=', False),
+            ], ['email'], groupby='email')
+        for email in attendees_emails:
+            attendee_partner = env['res.partner'].search([
+                ('email', '=ilike', email['email'])
+            ], limit=1)
+            if attendee_partner:
+                attendees = env['event.registration'].search(email['__domain'])
+                attendees.write({
+                    'attendee_partner_id': attendee_partner.id,
+                })
--- a/partner_event/__init__.py
+++ b/partner_event/__init__.py
@@ -1,2 +1,3 @@
 from . import models
 from . import wizard
+from .hooks import post_init_hook
--- a/partner_event/__manifest__.py
+++ b/partner_event/__manifest__.py
@@ -24,5 +24,6 @@
         'views/event_registration_view.xml',
         'wizard/res_partner_register_event_view.xml',
     ],
+    'post_init_hook': 'post_init_hook',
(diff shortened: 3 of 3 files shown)


partner_event/hooks.py
```diff
@@ -0,0 +1,21 @@
...
+# License AGPL-3 - See https://www.gnu.org/licenses/agpl-3.0.html
+from odoo import api, SUPERUSER_ID
+
+
+def post_init_hook(cr, registry):
+    """Preload proper attendee partner for existing registrations using
+       the same rules the module does"""
+    with api.Environment.manage():
+        env = api.Environment(cr, SUPERUSER_ID, {})
+        attendees = env['event.registration'].search([
```
pedrobaeza: Do a `read_group` by email for searching only once each email

partner_event/hooks.py
```diff
@@ -0,0 +1,21 @@
...
+            ('email', '!=', False),
+        ])
+        for attendee in attendees:
+            attendee_partner = Partner.search([
+                ('email', '=ilike', email)
+            ], limit=1)
+            if attendee_partner:
+                attendee.attendee_partner_id = attendee_partner
+            elif attendee.partner_id:
+                attendee.attendee_partner_id = attendee.partner_id
```
pedrobaeza: Not sure of this part. Can this be False?

chienandalu: Well, the only required field for an attendee is the event_id

pedrobaeza: Then better to leave it unfilled

chienandalu: ok then

partner_event/hooks.py
chienandalu: Change done

partner_event/hooks.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def post_init_hook(cr, registry):
+    """Preload proper attendee partner for existing registrations using
+       the same rules the module does"""
+    with api.Environment.manage():
+        env = api.Environment(cr, SUPERUSER_ID, {})
+        attendees_emails = env['event.registration'].read_group([
+            ('email', '!=', False),
+            ], ['email'], groupby='email', orderby='email')
```
pedrobaeza: Why ordering by `email`? Is this needed for something?

partner_event/hooks.py
```diff
@@ -0,0 +1,22 @@
...
+        for email in attendees_emails:
+            attendee_partner = env['res.partner'].search([
+                ('email', '=ilike', email['email'])
```
pedrobaeza, line 16: Why not `=`?

chienandalu: There's no need to be case sensitive.

partner_event/hooks.py
chienandalu: isn't it mandatory to have an `orderby` criteria?

partner_event/hooks.py
pedrobaeza: OK

partner_event/hooks.py
pedrobaeza: No, it's not mandatory.

chienandalu:
Changes done

pedrobaeza (approved):
Merging this directly as missing initialization option for not letting the module in unstable situation

PR #174, opened 2019-11-21
[FIX] website_event_snippet_calendar: Respect user lang

yajo:
These controllers were ignoring the `website_lang` cookie because they were not enabled for website. Thus, event results were downloaded in default language instead of the current one.

@Tecnativa TT20711


pedrobaeza (approved):
Please apply it also for v12

chienandalu (approved):

pedrobaeza:
Let's merge it for fixing the error

/ocabot merge patch

PR #175, opened 2019-11-25
[FIX] website_event_snippet_calendar: Respect user lang

yajo:
Forward-port of https://github.com/OCA/event/pull/174.

These controllers were ignoring the `website_lang` cookie because they were not enabled for website. Thus, event results were downloaded in default language instead of the current one.

@Tecnativa TT20711


yajo:
Since this is a fwport...

/ocabot merge patch

PR #177, opened 2020-01-28
Add webinar functionality to events [website_event_webinars]

chrisandrewmann:
Adds webinar functionality to events with:
-New field on events and event registrations titled 'Webinar ID/URL' [webinar_id]
Ability to store the ID/URL of a webinar and can be used to integrate i.e link to Zapier for Zoom integration
-New field on event titled 'Restrict ticket to logged-in user' [website_attendee_require_login]
Forces event to only allow ticket under logged-in email / name


website_event_webinars/.gitignore
```diff
@@ -0,0 +1,352 @@
+## Ignore Visual Studio temporary files, build results, and
```
emagdalenaC2i: Remove file

emagdalenaC2i (commented):
Please, could you exxplain the differences with website_event_require_login

website_event_webinars/.gitignore
chrisandrewmann: Hi emagdalenaC2i. 
Just to confirm that this module compliments the website_event_require_login, does not replace it.
website_event_require_login forces login before allowing registration for the event.
My module website_event_webinars goes further to also FORCE that when registering for a ticket for an event, it can only be registered under the logged-in name and email address.
Without it, you may free-type any other person's name and email into the ticket registration box. For "invite only" webinar events, you may need to vet people beforehand and therefore not allow anonymous ticket submissions.
Hope that helps.
Chris

emagdalenaC2i: Ok, thanks

chienandalu (commented):
In a light review, I feel this should be modeled with partner_event as a base for create the proper users portal and send them credentials.

PR #179, opened 2020-04-07
[MOD] 12.0 Replace code using system lang by babel dates translations

llacroix:
The big issue with this kind of code is that if you have a user with a
certain locale being used, it will override the locale system wide and
the next request on the server may have the dates formatted with
strftime in a different language than expected...

The other issue is that some systems do not provide a set of actual
locales so on a server running in docker without manually modifying
a dockerfile to install a set of locales, it may fail to switch
the locale to a non existing one. It will just fail to compute the field.

Babel is already a dependency of odoo and it's a bit strange to not use it.

This probably can be ported to other versions of event repository.

### DIFF
--- a/event_session/models/event_session.py
+++ b/event_session/models/event_session.py
@@ -4,7 +4,51 @@
 
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
-from locale import setlocale, LC_ALL
+import babel
+from babel.dates import format_datetime, format_time
+
+
+def get_locale(env):
+    """
+    Get the locale from the environment as done in ir.qweb.field
+
+    https://github.com/odoo/odoo/blob/12.0/odoo/addons/base/models/ir_qweb_fields.py#L235
+    """
+    lang = env['ir.qweb.field'].user_lang()
+    locale = babel.Locale.parse(lang.code)
+    return locale
+
+
+def localized_format(value, formats, locale):
+    """
+    Return a string separated by spaces of the formatted
+    value based on the locale passed as argument.
+    """
+    values = [
+        dt_format(value, locale)
+        for dt_format in formats
+    ]
+    return " ".join(values)
+
+
+def time_format(dt_format):
+    """
+    Returns a callable that will format a datetime with a locale
+    using the format_time method of babel.
+    """
+    def wrap(value, locale):
+        return format_time(value, dt_format, locale=locale)
+    return wrap
+
+
+def datetime_format(dt_format):
+    """
+    Returns a callable that will format a datetime with a locale
+    using the format_datetime method of babel.
+    """
+    def wrap(value, locale):
+        return format_datetime(value, dt_format, locale=locale)
+    return wrap
 
 
 class EventSession(models.Model):
@@ -98,20 +142,35 @@ class EventSession(models.Model):
     @api.multi
     @api.depends('date_begin', 'date_end')
     def _compute_name(self):
-        setlocale(LC_ALL, locale=(self.env.lang, 'UTF-8'))
+        locale = get_locale(self.env)
         for session in self:
             if not (session.date_begin and session.date_end):
                 session.name = '/'
                 continue
             date_begin = fields.Datetime.from_string(
-                session.date_begin_located)
-            date_end = fields.Datetime.from_string(session.date_end_located)
-            dt_format = '%A %d/%m/%y %H:%M'
-            name = date_begin.strftime(dt_format)
+                session.date_begin_located
+            )
+            date_end = fields.Datetime.from_string(
+                session.date_end_located
+            )
+
+            dt_formats = [
+                datetime_format("EEEE"),
+                datetime_format("short")
+            ]
+
+            name = localized_format(date_begin, dt_formats, locale)
+
             if date_begin.date() == date_end.date():
-                dt_format = '%H:%M'
-            name += " - " + date_end.strftime(dt_format)
-            session.name = name.capitalize()
+                dt_formats = [
+                    time_format("short")
+                ]
+
+            name = "%s - %s" % (
+                name,
+                localized_format(date_end, dt_formats, locale)
+            )
+            session.name = name
 
     def _session_mails_from_template(self, event_id, mail_template=None):
         vals = [(6, 0, [])]
--- a/event_session/__manifest__.py
+++ b/event_session/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     'name': 'Event Sessions',
-    'version': '12.0.1.0.1',
+    'version': '12.0.1.0.2',
     'author': 'Tecnativa, '
               'Odoo Community Association (OCA)',
     "license": "AGPL-3",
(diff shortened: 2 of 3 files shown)


event_session/models/event_session.py
```diff
@@ -98,20 +98,37 @@ class EventSession(models.Model):
...
+                format_datetime(value, format, locale=self.env.lang)
+                for value in formats
+            ]
+            return " ".join(values)
+
         for session in self:
             if not (session.date_begin and session.date_end):
                 session.name = '/'
                 continue
+
```
chienandalu: Please, don't intruduce unnecessary blank lines

chienandalu (changes requested):
Tests are failing.

Can you describe the issue enforcing this change?

event_session/models/event_session.py
llacroix: ok

llacroix:
There are multiple issues here:
1. setLocale will fail with an exception if the locale isn't present on the host
2. Locale changes are system wide.. So after the method finishes, the locale would remain to the locale set in the compute call.

Here's a small snippet to understand:

```
from datetime import datetime
from locale import setlocale, LC_ALL

def print_date(locale=None):
    if locale:
        setlocale(LC_ALL, locale)
    date = datetime.now()
    print(date.strftime('%A'))

print_date('en_CA')                                                                                                                                                                                                                           
print_date('fr_CA')
print_date()
```

The issue is that you have a user in a certain locale triggering the recompute and switching the current worker locale... then somewhere you'll start to have things rendered in different languages without knowing exactly why.
One easy example where this can lead to very difficult issue to fix/figure out is when you try to parse things that are locale dependant. For example, using the above snippet you'll see how keeping the locale changed can lead to breaking other part of code.

```
from datetime import datetime
from locale import setlocale, LC_ALL

def print_date(locale=None):
    if locale:
        setlocale(LC_ALL, locale)
    date = datetime.now()
    print(date.strftime('%A'))

def parse_time(text):
    return datetime.strptime(text, '%A')

print_date('en_CA')
print(parse_time('Wednesday'))
print_date('fr_CA')
print(parse_time('Wednesday'))                                                                                                                                                                                                                
print_date()
```
It won't be possible to parse "Wednesday" later because it expect "mercredi". There are probably countless of other reasons why changing the locales is a bad idea.

Babel is already a dependency of odoo and that's the method used in t-field parsing on Qweb views. 

    <span t-field="obj.date_object" t-options="{'format': 'EEEE'}" /> 

Is the right way to display the weekday in QWeb, using t-esc + strftime will lead to have dates rendered in a different language than the user's session language. It only make sense to use the same method inside odoo if you want to render the session name in the partner's language. 

I'll fix the errors, it was a quick patch I made yesterday.

event_session/models/event_session.py
```diff
@@ -98,20 +98,36 @@ class EventSession(models.Model):
     @api.multi
     @api.depends('date_begin', 'date_end')
     def _compute_name(self):
-        setlocale(LC_ALL, locale=(self.env.lang, 'UTF-8'))
+        def format_field(value, formats):
```
NL66278: If you make this into an independent function, you could put some unit tests for this testing/demonstrating usage with at least two different languages.

llacroix: Yes not a bad idea, I already moved it out of the function so it can be tested accordingly. I also added a new method to get the locale based on the way ir.qweb.fields does to prevent a bug where self.env.lang is None and apparently odoo default to en_US when self.env.lang is None.

llacroix:
Finally! It needs some tests for the new methods and probably some test showing that changing the language works with different languages.

NL66278 (approved):
:+1: LGTM Code Review

llacroix:
Hmm ah... this is a bit deceiving I just found why the test fails:

```
class UserLang(models.Model):
    ...
    @tools.ormcache('code')
    def _lang_get_id(self, code):
        return (self.search([('code', '=', code)]) or                                                                                                                                                                                         
                self.search([('code', '=', 'en_US')]) or
                self.search([], limit=1)).id

    @api.model
    @api.returns('self', lambda value: value.id)
    def _lang_get(self, code):
        return self.browse(self._lang_get_id(code))
```
So the tests fail because the lang isn't installed.. so it has active=False and search will strip it out.. The code actually work in odoo but doesn't let test it properly with uninstalled languages..

llacroix:
Should be good now, I probably won't add more tests. It's testing that the code doesn't work for translations in different locales based on the context language. 

There are few limitation to the module that I noticed while doing that. It wasn't clear from the start that the name of the session is computed with store=True... What it means is that the translation is fixed... It also means that if a module want to display session.name on the website for example, the translation will be wrong for some users as the computed won't be recomputed for different partners.
For website and displaying it's not a big issue because we can simply display the date_begin, date_end properly formatted in a qweb template.

Language that can be used are only the active languages so even thought babel can handle most languages existing.. in odoo, it won't be able to use a language that isn't enabled... for example if a language isn't active it will default to en_US.. 

Locales may differ from one platform to an other, for example on my computer for whatever reason, en_CA is rendered as p.m. instead of PM on the servers here (might be distribution independent).. 

When executing the module inside docker, I also have the same behaviour of PM instead of p.m. on my host. So there is probably something going on there. 

Tests aren't going to reliably work on all platforms the same way but if for some reason the platform change and locales are indirectly gives a different output, the change will be caught by the tests failing. So unless people report problems, I'd be ok to leave the code as is now... But one thing that could be done later is to test with a direct call to babel and compare both cases but if the platform behaviour changes it will silently pass the tests as long as the same locale is passed. Thought it could raise a good issue.

pedrobaeza:
Wow, it seems you have dug a lot.

Can you please squash commits?

llacroix:
Yes, I can do that

llacroix:
I also upgraded version to 1.0.2 for the module

llacroix:
@chienandalu ?

chienandalu (approved):
Ok, thanks for the work and your explanation!

pedrobaeza:
@sbidoul ocabot seems stalled. I have issued the command in 2 PRs and it's not reacting.

rvalyi:
> @sbidoul ocabot seems stalled. I have issued the command in 2 PRs and it's not reacting.

For the record it happenned to us last week in l10n-brazil and we ended up merging manually. Subsequent PRs worked again. In our case there is a suspicion we had some msgmerge (po/weblate stuff) command failing and we had no feedback about it. Well Travis was all green (not testing msgmerge) and in the next PRs we had some msgmerge error that obviously was already present in our 1st PR were ocabot hanged. Well it might not be the related but again that was the only thing suspicious.

sbidoul:
Looks like the bot was sill busy with the nightly jobs and was queuing requests. I'll need to investigate that.

pedrobaeza:
OK, thanks. Finally it seems it has started.

pedrobaeza:
But the CIs of the merge branch has finished and no merge is done...

sbidoul:
A little bit of patience. All this is asynchronous

sbidoul:
The branch is still there and a new test is being run following a .pot udpate (https://github.com/OCA/event/commits/12.0-ocabot-merge-pr-179-by-pedrobaeza-bump-patch). Everything operating normally.

pedrobaeza:
OK, sorry for the impatience.

sbidoul:
No worries @pedrobaeza. The bot could probably report a little bit more feedback.

sbidoul:
@rvalyi when this happens don't hesitate to ping me.

llacroix:
Thanks!

PR #181, opened 2020-04-16
12.0 website event session

llacroix:
**Website Event Session Sale**

This module allow users to display website event session on their e-commerce website.
Sessions are filtered by dates in the future by default using properly set ir.rules (time based).

This patch also fixes some issues in the module event_session_sale that expect a sale.order.line to be always defined on the registrations... In reality the website flow doesn't provide a sale.order.line at the moment the registration is created. For that reason, it's best to check if a sale.order.line is defined than use it directly with `obj[field]`.
Also since it's possible to directly pass the session_id in the registration, this code change check if the session_id is set and do not override it if it is. 

As a result, this code works with the website and the backend.


llacroix:
Any idea how to test the controllers ?

event_sale_session/models/event_registration.py
```diff
@@ -8,8 +8,8 @@ class EventRegistration(models.Model):
 
     @api.model
     def _prepare_attendee_values(self, registration):
-        data = super(EventRegistration, self)._prepare_attendee_values(
-            registration)
-        session_id = registration['sale_order_line_id'].session_id.id
-        data.update({'session_id': session_id})
+        data = super(EventRegistration, self)._prepare_attendee_values(registration)
+        if not data.get('session_id') and 'sale_order_line_id' in registration:
```
chienandalu, line 12: It's preferable to make this kind of changes in a separate PR and document the motivations.

website_event_session_sale/README.rst
```diff
@@ -0,0 +1,24 @@
+=====================================================
+Website Event Custom
+=====================================================
+
+Customize event registration made from the website
```
chienandalu: You should follow OCA's readme conventions: https://github.com/OCA/maintainer-tools/tree/master/template/module/readme

website_event_session_sale/views/event_templates.xml
```diff
@@ -0,0 +1,46 @@
...
+    id="registration_attendee_details"
+    inherit_id="website_event.registration_attendee_details"
+    priority="10"
+    >
+    <xpath expr="//*[hasclass('modal-body')]/div/t[last()]" position="after">
+      <t t-if="event.session_ids">
+        <h4 class="o_page_header mt8"><strong>Session</strong></h4>
+        <div>
+          <div class="col-lg-6">
+            <select t-attf-name="#{counter}-session_id" required="true">
```
chienandalu:
```suggestion
            <select t-attf-name="#{counter}-session_id" required="true" class="form-control">
```

chienandalu (changes requested):
- In the formal part of the code, you should follow OCA's guidelines: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst
- Testing UI: you've got plenty of examples in https://github.com/OCA/e-commerce

pedrobaeza:
For the module name, I would say `website_event_sale_session` for using the prefix over the dependent module (`website_event_sale`)

website_event_session_sale/README.rst
llacroix: Ok

PR #182, opened 2020-04-23
[FIX] website_event_excerpt_img: fix some bootstrap 4 styles

yajo:
Some styles were not properly migrated. Fixing here. Pretty obvious fixes.

@Tecnativa TT17632


pedrobaeza:
OK, fast-tracking then

/ocabot merge minor

PR #184, opened 2020-04-24
[IMP] website_event_filter_selector: make prettier

yajo:
Use `custom-select` in `<select>` elements instead of just `form-control`. This way BS4 can style the dropdowns more freely and they look more consistent with the rest of the site.

Before:

After:

@Tecnativa TT17632


pedrobaeza (approved):

chienandalu (approved):

yajo:
Merge?

PR #188, opened 2020-09-07
[13.0] [MIG] partner_event: Migration to v13.0

victoralmau:
@Tecnativa TT25337

Please @pedrobaeza  review it


chienandalu (commented):
¿no migration commit?

victoralmau:
Yes, sorry, it was an error and it seems that everything has been mixed in the same commit. 
I take note of it so that it doesn't happen anymore.

pedrobaeza:
You need to extract them. We will see it how tomorrow.

victoralmau:
Sorry, the commit with the migration changes is already there.

partner_event/__manifest__.py
```diff
@@ -4,17 +4,18 @@
...
 
 {
     "name": "Link partner to events",
-    "version": "12.0.1.0.1",
+    "version": "13.0.1.0.0",
     "category": "Marketing",
     "author": "Tecnativa," "Odoo Community Association (OCA)",
     "website": "https://www.tecnativa.com",
-    "development_status": "Production/Stable",
+    "development_status": "Alpha",
```
pedrobaeza: Why are you changing this status? It should be the same. Check https://odoo-community.org/page/module-maturity-levels

partner_event/models/event_event.py
```diff
@@ -2,6 +2,7 @@
 # Copyright 2015 Tecnativa S.L. - Javier Iniesta
 # Copyright 2016 Tecnativa S.L. - Antonio Espinosa
 # Copyright 2016 Tecnativa S.L. - Vicent Cubells
+# Copyright 2020 Tecnativa S.L. - Víctor Martínez
```
pedrobaeza: If no change on the file, you shouldn't add your copyright

partner_event/models/event_registration.py
```diff
@@ -35,19 +36,17 @@ def create(self, vals):
             attendee_partner = Partner.search([("email", "=ilike", email)], limit=1)
             event = Event.browse(vals["event_id"])
             if attendee_partner:
-                vals["name"] = vals.setdefault("name", attendee_partner.name)
-                vals["phone"] = vals.setdefault("phone", attendee_partner.phone)
+                for key in ["name", "phone", "mobile"]:
+                    if attendee_partner[key]:
```
pedrobaeza: This code is not equivalent, as it's overwriting possible existing values in `vals`, which previous code was not doing.

partner_event/readme/CONTRIBUTORS.rst
```diff
@@ -6,3 +6,4 @@
 * David Vidal <david.vidal@tecnativa.com>
 * Rafael Blasco <rafael.blasco@tecnativa.com>
 * Anil Kesariya <anil.r.kesariya@gmail.com>
+* Víctor Martínez <victor.martinez@tecnativa.com>
```
pedrobaeza: We have changed the way to reflect contributors here with:

- No emails, for avoiding spam
- Grouping all our contributors under a general point with company's website. Example: https://raw.githubusercontent.com/OCA/account-invoicing/13.0/account_invoice_validation_queued/readme/CONTRIBUTORS.rst

partner_event/__manifest__.py
chienandalu: Why is `Alpha` now? :thinking:

pedrobaeza (changes requested):

partner_event/models/event_event.py
chienandalu: If no relevant changes in the logic of the code are done, you shouldn't add your self to author lines (only in contributors on the README.rst)

partner_event/models/event_registration.py
```diff
@@ -0,0 +1,68 @@
...
+    )
+
+    def _prepare_partner(self, vals):
+        return {
+            "name": vals.get("name") or vals.get("email"),
+            "email": vals.get("email", False),
+            "phone": vals.get("phone", False),
+        }
+
+    @api.model
```
chienandalu: It'd be nice to pass it to `@api.model_create_multi`

partner_event/models/event_registration.py
```diff
@@ -35,19 +36,17 @@ def create(self, vals):
             attendee_partner = Partner.search([("email", "=ilike", email)], limit=1)
             event = Event.browse(vals["event_id"])
             if attendee_partner:
-                vals["name"] = vals.setdefault("name", attendee_partner.name)
-                vals["phone"] = vals.setdefault("phone", attendee_partner.phone)
```
chienandalu: Out of curiosity. Why didn't this work?

chienandalu (changes requested):
Thanks @victoralmau Some remarks:

partner_event/__manifest__.py
victoralmau: Sorry, I thought that when doing the migration you should change.
I write it down for a future that must be left.

partner_event/models/event_event.py
victoralmau: Sorry, maybe some change was made initially and not in the end and for that reason it would be added.
It is already removed.

partner_event/models/event_registration.py
victoralmau: Finally this part has changed because in V12 it did not work (you could not define a value of something that already had a False value by default in the create function).

I have changed the create function because if there is a partner and we associate it, we must force in that record to save the fields of the record (name, phone, mobile).

partner_event/readme/CONTRIBUTORS.rst
victoralmau: The file has been changed.

partner_event/__manifest__.py
victoralmau: https://github.com/OCA/event/pull/188#discussion_r484865473

partner_event/models/event_event.py
victoralmau: https://github.com/OCA/event/pull/188#discussion_r484865941

partner_event/models/event_registration.py
victoralmau: https://github.com/OCA/event/pull/188#discussion_r484867623

partner_event/models/event_registration.py
victoralmau: Some example addon that uses `@api.model_create_multi` to know what things should be changed and / or how to change them?

chienandalu: Sure :)

https://www.odoo.com/documentation/13.0/reference/orm.html#odoo.api.model_create_multi
A simple example:
https://github.com/odoo/odoo/blob/054d4bc6bc219bcc6b0a64265e8d7e9c7423dbc8/addons/uom/models/uom_uom.py#L96-L102

partner_event/models/event_registration.py
```diff
@@ -35,19 +36,17 @@ def create(self, vals):
...
+                vals["name"] = attendee_partner.name
```
chienandalu: Sorry, but the change is incorrect, as it changes the expected behavior. For example, for the case that the same email address is used for several attendees (a real possibility when  the registration is made from the website) we'd loose that information.

Do these keys come in the `vals` dictionary when creating the attendee. If so, then use coercion to default to the proper value. Something like

```python
vals["name"] = not vals.get("name") and attendee_partner.name or vals.get("name")
```

victoralmau: Thanks for the example, uploaded changes.

partner_event/models/event_registration.py
victoralmau: Thanks for the example, uploaded changes.

chienandalu (approved):
Ok :) :+1:

partner_event/models/event_registration.py
```diff
@@ -25,29 +26,43 @@ def _prepare_partner(self, vals):
...
+                    values["phone"] = (
+                        not values.get("phone")
+                        and attendee_partner.phone
+                        or values.get("phone")
+                    )
+                    values["mobile"] = (
+                        not values.get("mobile")
+                        and attendee_partner.mobile
+                        or values.get("mobile")
+                    )
```
pedrobaeza: Simplify this expression:
```suggestion
                    for field in {"name", "phone", "mobile"}:
                        vals[field] = vals.get(field) or attendee[field]
```

partner_event/models/event_registration.py
```diff
@@ -25,29 +26,43 @@ def _prepare_partner(self, vals):
...
+                elif event.create_partner:
+                    # Create partner
+                    attendee_partner = Partner.sudo().create(
+                        self._prepare_partner(values)
+                    )
+                values["attendee_partner_id"] = attendee_partner.id
+
```
pedrobaeza: Don't put empty lines inside a method

pedrobaeza (changes requested):

partner_event/models/event_registration.py
victoralmau: Thanks, it's already removed now.

partner_event/models/event_registration.py
victoralmau: Ok it's already simplified

pedrobaeza (approved):

PR #200, opened 2020-12-12
[14.0] dotfiles update needs manual intervention

sbidoul:
Dear maintainer,

After updating the dotfiles, `pre-commit run -a`
fails in a manner that cannot be resolved automatically.

Can you please have a look, fix and merge?

Thanks,


.copier-answers.yml
```diff
@@ -1,11 +1,12 @@
...
-_commit: v1.0.2
+_commit: v1.1.1
 _src_path: gh:oca/oca-addons-repo-template
 dependency_installation_mode: PIP
+generate_requirements_txt: true
 include_wkhtmltopdf: false
 odoo_version: 14.0
 rebel_module_groups: []
-repo_description: "TODO: add repo description."
+repo_description: 'TODO: add repo description.'
```
pedrobaeza, line 9: @sbidoul It's curious that this is changed, but afterwards pre-commit fails because of it. Maybe the template is incorrect and include single quote instead of double?

sbidoul: Yes the normal process is `copier -f update && pre-commit run -a` which automatically fixes these issues.
Here copier update created a `.rej` file so my script stopped on that and did not run pre-commit run -a.x

But yes, changing the quotes in the template can't hurt.

sbidoul: > But yes, changing the quotes in the template can't hurt.

Ah, no this file is generated by copier itself, maybe we should add it to the pre-commit exclude list.

pedrobaeza: Or do it pre-commit friendly. cc @Yajo

sbidoul:
Anyway, I fixed manually.

.copier-answers.yml
yajo: The real problem is the `.rej` file below, as @sbidoul said. There's a merge conflict, thus that requires manual intervention.

This problem with quotes would be fixed with pre-commit automatically after the conflicts are fixed, so not much to do here.

The YAML is dumped with PyYAML, which doesn't always dump the prettiest YAML, so the easiest fix is anyways to run pre-commit later.

yajo:
@sbidoul I'm curious. This PR is created by your updating script? 🤔

sbidoul:
@yajo yes: https://github.com/OCA/maintainer-tools/pull/486

PR #210, opened 2021-04-05
[13.0][MIG] event_session

dalonsod:
Standard migration. Depends on #194


dalonsod:
@chienandalu @sergio-teruel could you review? Thanks!

event_session/models/event_session.py
```diff
@@ -114,9 +109,9 @@ class EventSession(models.Model):
         store=True,
     )
     seats_available_pc = fields.Float(
-        string="Full %", readonly=True, compute="_compute_seats",
+        string="Full %", store=True, readonly=True, compute="_compute_seats",
```
chienandalu: Is there a motivation to change this field to stored?

dalonsod: Yes, there was. Runbot failed with a warning. because `_compute_seats` was used for stored fields but `seats_available_pc`. The error was "inconsistent compute_sudo for computed fields", a v13 feature I think. Maybe an alternative for this field should be forcing `compute_sudo=True` instead of making it stored, and warning should be avoided too. What do you think?

pedrobaeza: Please use `compute_sudo=True` then.

dalonsod: Done!

chienandalu (approved):
Thanks @dalonsod

PR #214, opened 2021-05-12
[ADD][13.0] event_type_multicompany

ivantodorovich:
Adds a `company_id` field in `event.type` model and the corresponding security rule


simahawk (approved):
Not blocking: would be nice to have a test.
Question: would it be worth to propose this to odoo/master?

santostelmo (approved):

rousseldenis:
> Question: would it be worth to propose this to odoo/master?

Indeed, would be nice

cvinh:
Hello
Maybe it would be nice in https://www.github.com/oca/multi-company ?
Best regards

simahawk:
> Hello Maybe it would be nice in https://www.github.com/oca/multi-company ? Best regards

this belong to events mainly so I think the current position is ok.

Regarding odoo/master: let's propose for v16 :rofl:

PR #215, opened 2021-05-19
[13.0][ADD] pos_event_sale : Sell events from the Point of Sale

ivantodorovich:
New module to sell events from Point of Sale

---

I wasn't sure where to push it, but ended up choosing this repo over OCA/pos because it's more specific IMO.

Also I imagine it's possible we develop glue modules with other modules in this repo (ie: `pos_event_session_sale`), and it's better to have them all in one repo.

---

add_event_button

event_calendar

ticket_selector


ivantodorovich:
ping @OCA/pos-maintainers as they may not be watching this repo but could be more qualified to review

legalsylvain:
Big feature !
Thanks !

legalsylvain:
Hi @ivantodorovich. Thinking about your question regarding repo : 

- In a business logic point of view, both repos are valid, because it's about PoS & about events

- In a visibility point of view, both repos are valid, because people want that feature, they will look in OCA/pos and in OCA/event

- but in a developper point of view, I think that OCA/pos is the right choice : all the difficulties of your module are PoS (JS, css, etc...). there are only few lines of "event". if all the ``pos_xxx`` modules are in OCA/xxx repositories, it will not help to develop new modules, or improve existing ones. Exemple : if tomorrow, I want to develop a module ``pos_purchase`` to create purchase order via PoS, I will take inspiration from other modules I find. If ``pos_event`` is in event, I will not find it. if ``pos_order_to_sale_order`` had been in OCA/sale-workflow I would have not find it also.
I think that all the PoS module should be in OCA/pos to help developpers to copy good technics from a module to another. (how to overload correctly PoS, how to write tours, etc...)

what do you think ? 

No code review / No test.

ivantodorovich:
Yes from a developer perspective I agree, it's better to have it in OCA/pos.
I've just created the new PR https://github.com/OCA/pos/pull/667

Closing this one

Thanks!

PR #216, opened 2021-05-24
[FIX] partner_event: create registrations from attendees

chienandalu:
If we launch the "Register to event" from the attendees list we should
evaluate the active model and extract the proper partners from the
attendees lists.

cc @Tecnativa TT29857


pedrobaeza (approved):

pedrobaeza:
Fast-tracking the obvious patch:

/ocabot merge patch

fw-port if proceed to v13/v14

PR #217, opened 2021-05-28
[13.0][ADD] event_sale_portal : Module to display confirmed event registrations in the Sale Order portal page.

ivantodorovich:
Displays the event registrations in the sale order portal page, and allows the user to download the event badges.

portal

### DIFF
--- a/event_sale_portal/controllers/portal.py
+++ b/event_sale_portal/controllers/portal.py
@@ -1,0 +1,60 @@
+from odoo import http
+from odoo.exceptions import AccessError, MissingError
+from odoo.http import content_disposition, request
+
+from odoo.addons.portal.controllers.portal import CustomerPortal
+
+
+class CustomerPortal(CustomerPortal):
+    @http.route(
+        [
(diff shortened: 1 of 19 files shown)


event_sale_portal/controllers/portal.py
```diff
@@ -0,0 +1,60 @@
...
+                r.state == "open"
+                and r.event_id.id == event_id
+                and r.event_id.portal_badge_download
+            )
+        )
+        if not registrations:
+            return request.redirect("/my")
+        # Check if registration_id belongs to the sale order
+        if reg_id is not None:
+            if reg_id not in registrations.ids:
```
ajaniszewska-dev: not blocking:
```suggestion
        if reg_id and reg_id not in registrations.ids:
```

ivantodorovich: It's not the same though. See [line 42](https://github.com/OCA/event/pull/217/files#diff-e7b9aa0c19b7535023d1d615dd8cfc93c51f6f36b2f2d2480f0f798682e43d27R42)

PR #218, opened 2021-05-28
[13.0][ADD] event_registration_lang

ivantodorovich:
Module that allows to set an attendee language


pedrobaeza:
Isn't better to go with `partner_event`, and then you have the language there?

ivantodorovich:
`partner_event` is creating a partner for each registration, and it does [some "magic" looking up partners with the same email](https://github.com/OCA/event/blob/8a4cdbc757c269eaf11406f63ae841e03bb7424a/partner_event/models/event_registration.py#L35) that is not always wanted.

In the project I'm working on, the core behaviour + this module is enough

ajaniszewska-dev (approved):

chienandalu (commented):
Code review :+1:

PR #219, opened 2021-05-31
[13.0][FIX] event_session: Error when seats_max is 0

CarlosRoca13:
When we create a session with seats_max = 0, we get an error when we try to view it.

Doing this changes we achieve to avoid None values on the fields seats_available and seats_available_pc.

cc @Tecnativa TT29930

### DIFF
--- a/event_session/models/event_session.py
+++ b/event_session/models/event_session.py
@@ -249,6 +249,10 @@ class EventSession(models.Model):
                 session[state_field[res["state"]]] += res["__count"]
         # compute seats_available
         for session in self:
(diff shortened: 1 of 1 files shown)


CarlosRoca13:
@chienandalu @pedrobaeza review this, please

CarlosRoca13:
Example of error before the patch 

image

pedrobaeza (approved):

pedrobaeza:
Fast-tracking obvious patch:

/ocabot merge patch

chienandalu (approved):

PR #220, opened 2021-05-31
[12.0][FIX] event_registration_multi_qty: Force quantity despite default

joao-p-marques:
After https://github.com/odoo/odoo/commit/a4d50b468c8f37155048e2c7bb0c816f45d27a95, the passed value is not taken into account when a default value is set on the field.
This forces the update to the quantity.

cc @Tecnativa @chienandalu

### DIFF
--- a/event_registration_multi_qty/models/event.py
+++ b/event_registration_multi_qty/models/event.py
@@ -77,3 +77,12 @@ class EventRegistration(models.Model):
                     _('You can not add quantities if you not active the'
                       ' option "Allow multiple attendees per registration"'
                       ' in event'))
+
+    @api.model
+    def _prepare_attendee_values(self, registration):
+        res = super()._prepare_attendee_values(registration)
+        # Passed fields are not taken into account if a default is set,
+        # so we need to force this
+        if "qty" in registration:
+            res.update({"qty": registration.get("qty")})
+        return res


event_registration_multi_qty/models/event.py
```diff
@@ -77,3 +77,13 @@ def _check_attendees_qty(self):
                     _('You can not add quantities if you not active the'
                       ' option "Allow multiple attendees per registration"'
                       ' in event'))
+
+    @api.model
+    def _prepare_attendee_values(self, registration):
+        res = super()._prepare_attendee_values(registration)
+        # Passed fields are not taken into account if a default is set.
+        # We can force this with a specific context
+        qty = self.env.context.get("default_qty")
```
chienandalu: Why not to check `registration.get("qty")` instead?

joao-p-marques: Ok, it's just that, in that case, we would be completely skipping Odoo's mechanism applied in the mentioned commit. Maybe that was a necessary fix for some use case and we might run into problems in the future? This way, we would avoid changing a lot of the current behavior...

chienandalu: It's not a change of the behavior. It's expected that this field doesn't default unless it's not informed. So if a method aks to register 5 attendees who are we to state the contrary :sweat_smile:

joao-p-marques: Ok, makes sense :smiley: Changes made!

joao-p-marques:
@pedrobaeza @victoralmau can you review this, please?

pedrobaeza (approved):

victoralmau (approved):

pedrobaeza:
@chienandalu good for you?

chienandalu (approved):

pedrobaeza (commented):
One thing before merging: is this field in core? If so, please add tests. If not, you should put an explanation as comment.

joao-p-marques:
No, the field is added in this module. Is the current comment enough to explain it?

pedrobaeza:
Why then you check about `if "qty" in registration:`? If it's a record, it will always be `True`.

chienandalu:
It's not a record, but a dictionary. Look: https://github.com/odoo/odoo/blob/a4d50b468c8f37155048e2c7bb0c816f45d27a95/addons/event/models/event.py#L439-L442

@joao-p-marques Maybe a link to the original method would ease the understanding of this change

pedrobaeza:
OK, the variable name is very bad (it should be `registration_vals` or so). And what about the test exercising this?

joao-p-marques:
> the variable name is very bad

Yes, but it comes from core :man_shrugging: 

> And what about the test exercising this?

Well, you said above to add tests if it the field came from core, which it does not. Do you still think it is worth it?

pedrobaeza:
Yes, I didn't express myself correctly. I mean that it comes with the dependency chain, not in extra module and this is a hack for avoiding a glue module. If this problem has arisen, a regression test would avoid to appear again in the future. Is it very complicated?

joao-p-marques:
Ok. Well, I guess I can add a simple test that calls the function and asserts the value is actually changed

pedrobaeza:
I'm afraid that doesn't serve. Tests should check real flows, not direct function calls. If not, in an upper version or future code revision, implementation may change, but the test will be correct.

joao-p-marques:
Well, then maybe @chienandalu can shed some light into what an appropriate test would be. Quite honestly, I didn't find this issue in UI, but in integration tests with another module, so that was the case I tested...

chienandalu:
The problem here is that although `_prepare_attendee_values` is defined in `event` is only used in core with `event_sale` and `website_event_sale`. I don't see how can we test it without adding one of those dependencies. That's why this error raised in a customer custom code test, which has all these dependencies joined althogether

pedrobaeza:
Ok, let's assume this and go:

/ocabot merge patch

PR #228, opened 2021-06-28
[ADD] module event_sale_reserve_registration

damdam-s:


event_sale_reserve_registration/views/sale_view.xml
```diff
@@ -0,0 +1,40 @@
...
+  Copyright 2021 Camptocamp SA
+  License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+-->
+<odoo>
+    <record id="sale_order_view_form" model="ir.ui.view">
+        <field name="name">sale.order.form.inherit.event.sale</field>
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale.view_order_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//header/button[@name='action_draft']" position="after">
```
ivantodorovich: Not blocking, just suggesting

I tend to avoid `xpath` when it isn't necessary, as IMO it's more readable to do this:

```suggestion
            <button name="action_draft" position="after">
```

event_sale_reserve_registration/views/sale_view.xml
```diff
@@ -0,0 +1,40 @@
...
+                    type="object"
+                    string="Reserve attendees"
+                />
+                <button
+                    name="confirm_reserved_registrations"
+                    type="object"
+                    string="Confirm attendees"
+                />
+            </xpath>
+            <xpath expr="//div[@name='button_box']" position="inside">
```
ivantodorovich: Following up on my previous comment, this also works.
It's just my personal opinion though :smiley: 
```suggestion
            <div name="button_box" position="inside">
```

event_sale_reserve_registration/models/sale_order.py
```diff
@@ -0,0 +1,49 @@
...
+                action["views"] = form_view + [
+                    (state, view) for state, view in action["views"] if view != "form"
+                ]
+            else:
+                action["views"] = form_view
+            action["res_id"] = registrations.id
+        else:
+            action = {"type": "ir.actions.act_window_close"}
+
+        return action
```
ivantodorovich: I'm not sure `action_view_attendees`, `attendees_count`  and `event_registration_ids` should be in this module.
Seems generic enough to be in a separate module.

In fact IMO it's something odoo shuold've done in core

event_sale_reserve_registration/models/sale_order.py
```diff
@@ -0,0 +1,49 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import api, fields, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def do_registrations(self):
```
ivantodorovich: `do_registrations` seems too generic for a model like `sale.order`.
I'd go for something like this

```suggestion
    def update_event_registrations(self):
```

event_sale_reserve_registration/models/sale_order.py
```diff
@@ -0,0 +1,49 @@
...
+        for order in self:
+            order.order_line._update_registrations(
+                confirm=False, cancel_to_draft=False, registration_data=None
+            )
+
+        return True
+
+    def confirm_reserved_registrations(self):
```
ivantodorovich:
```suggestion
    def confirm_event_registrations(self):
```

ivantodorovich (approved):

PR #229, opened 2021-08-03
[FIX] website_event_filter_selector: ignore events from other websites

yajo:
To reproduce the problem:

1. Create an event on website 2.
2. Do not publish it.
3. Go to website 1.

If you were logged in, that event still appeared as unpublished. It shouldn't happen because it belongs to other website, no matter whether it is published or not.

@Tecnativa TT29748


CarlosRoca13 (approved):

Tardo (approved):

pedrobaeza (approved):

pedrobaeza:
Please move it to v13

PR #230, opened 2021-08-05
[13.0][FIX] website_event_filter_selector: ignore events from other websites

yajo:
To reproduce the problem:

1. Create an event on website 2.
2. Do not publish it.
3. Go to website 1.

If you were logged in, that event still appeared as unpublished. It shouldn't happen because it belongs to other website, no matter whether it is published or not.

@Tecnativa TT29748 FW-port of https://github.com/OCA/event/pull/229

### DIFF
--- a/website_event_filter_city/__manifest__.py
+++ b/website_event_filter_city/__manifest__.py
@@ -11,6 +11,7 @@
     "Tecnativa, "
     "Onestein, "
     "Odoo Community Association (OCA)",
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

pedrobaeza:
Fast-tracking as straight fw-port:

/ocabot merge patch

yajo:
/ocabot merge patch

No need to wait, as this was approved already for v12.

PR #235, opened 2021-10-08
[V12][FIX]Check User ACL on registration to prevent ACL exception

flotho:
Use the same syntax as https://github.com/OCA/OCB/blob/12.0/addons/event/models/res_partner.py#L12 otherwise user that are not in the group are getting an ACL error when trying to read/write partners.

### DIFF
--- a/partner_event/views/res_partner_view.xml
+++ b/partner_event/views/res_partner_view.xml
@@ -16,11 +16,11 @@
     <field name="name">Partner Form with registrations</field>
     <field name="model">res.partner</field>
     <field name="inherit_id" ref="base.view_partner_form"/>
+    <field name="groups_id" eval="[(4, ref('event.group_event_user'))]"/>
     <field name="arch" type="xml">
         <div name="button_box" position="inside">
             <button name="%(partner_event.act_partner_registration)d"
                     type="action"
-                    groups="event.group_event_user"
                     class="oe_stat_button oe_inline"
                     help="Count of events with confirmed registrations."
                     icon="fa-id-card">
--- a/partner_event/views/res_partner_view.xml
+++ b/partner_event/views/res_partner_view.xml
@@ -20,6 +20,7 @@
         <div name="button_box" position="inside">
             <button name="%(partner_event.act_partner_registration)d"
                     type="action"
+                    groups="event.group_event_user"
                     class="oe_stat_button oe_inline"
                     help="Count of events with confirmed registrations."
                     icon="fa-id-card">
--- a/partner_event/models/res_partner.py
+++ b/partner_event/models/res_partner.py
@@ -21,8 +21,6 @@ class ResPartner(models.Model):
     @api.multi
     @api.depends('event_registration_ids')
     def _compute_registration_count(self):
-        if not self.user_has_groups('event.group_event_user'):
-            return
         for partner in self:
             partner.registration_count = len(
                 self.env["event.registration"].search([
--- a/partner_event/models/res_partner.py
+++ b/partner_event/models/res_partner.py
@@ -21,6 +21,8 @@ class ResPartner(models.Model):
     @api.multi
     @api.depends('event_registration_ids')
     def _compute_registration_count(self):
+        if not self.user_has_groups('event.group_event_user'):
+            return
         for partner in self:
             partner.registration_count = len(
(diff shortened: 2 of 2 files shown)


gaelTorrecillas (approved):

pedrobaeza:
Steps to reproduce the problem?

flotho:
Get a user who is in employee only
install this module
create a partner

pedrobaeza:
Then the solution is to put the group on the partner view that adds this field in the smart button, not this one.

flotho:
this was our first idea, we did it on all the forms,
we recheck and let you know

partner_event/views/res_partner_view.xml
```diff
@@ -20,6 +20,7 @@
         <div name="button_box" position="inside">
             <button name="%(partner_event.act_partner_registration)d"
                     type="action"
+                    groups="event.group_event_user"
```
pedrobaeza: I said in the view, not in the element. This still arises the error.

flotho:
@pedrobaeza , let me expose our analysis with @gaelTorrecillas 
* We have a user not in the event user group who is trying to read a partner and we have an ACL issue on registration event
* We added user envent group to ALL the partner view refrencing something like event or registr
* Then the user was able to read the partner
* He tried to create a partner and had this issue
image
* We re-made a deep check of all partner views possibly refrencing events and registrations and found nothing more so all the views were supposed to not display those computed fields.

So now, we were looking for a reason why a computed field referencing registration what throwing this issue and we exposed the field vie get. As you can see in attached, nothing is effectivley referencing such fields. So computed fields should not be called and then throw the exception.

[Fields_View_Get.txt](https://github.com/OCA/event/files/7311730/Fields_View_Get.txt)

By reading the OCA/OCB source we found that Odoo added a ACL check in the computed fields, I supposed that it was the origin of the issue. 

Maybe you could share with us your analysis tu help us undestand why the XML changes are not enough for solving this ACL.

thanks for your advices

pedrobaeza:
Have you tried what I have told you about the group in the view?

pedrobaeza:
It should be enough with:

```xml
<record model="ir.ui.view" id="view_partner_form_registrations">
    ...
    <field name="groups_id" eval="[(4, ref('event.group_event_user'))]"/>
        ...
```

partner_event/views/res_partner_view.xml
flotho: done

flotho:
@pedrobaeza we first tried directly by configuring it in the Odoo interface before proposing a PR and as it was not working by setting the group in the Odoo we searched in the python code the possible cause of this behaviour.

By reading the source here https://github.com/OCA/OCB/blob/12.0/addons/event/models/res_partner.py#L12 we suspected that Odoo has the same issue and didn't succeed in handling it by the way you proposed.
Even the way they deal with the view here https://github.com/OCA/OCB/blob/12.0/addons/event/views/res_partner_views.xml#L13 is indicating that they didn't succeed in handling the issue the way you propose.

My idea is that for an obscur reason, partner fields are "recomputed" on all event fields and that's why the view setup is not enough.

Could you share your point of view please?

pedrobaeza:
The reason by Odoo to do that ugly patch is the same: they are not using the right tool. They are putting the group in the element, not in the view. Please remove the other code patch in this PR, and if you try in runbot, you will see that it's working correctly.

flotho:
> ... Please remove the other code patch in this PR, and if you try in runbot, you will see that it's working correctly.

You mean python part ?
what if the field is called through API ?

flotho:
and for sure we'll try on runbot

pedrobaeza:
> You mean python part ? what if the field is called through API ?

Yes, Python part. There's no problem with API, unless you try to access specifically to that field, that you will receive the error if obviously you don't have permissions, but that's the expected behavior. Reading it with a value of 0 is totally incorrect.

flotho:
First test done with python code, no issue,

Let's try the next test without the python code

flotho:
rerunning the buil https://runbot.odoo-community.org/runbot/build/3521918 for testing

ivantodorovich (approved):

pedrobaeza:
Is this needed in upper versions?

PR #236, opened 2021-10-15
[ADD] module website_event_sale_hide_ticket

damdam-s:
new module to be able to hide event tickets from the website

### DIFF
--- a/website_event_sale_hide_ticket/__manifest__.py
+++ b/website_event_sale_hide_ticket/__manifest__.py
@@ -1,0 +1,16 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+{
+    "name": "Website Event Sale Hide Ticket",
+    "summary": "Allow to hide event ticket from the website",
+    "version": "13.0.1.0.0",
+    "development_status": "Beta",
+    "category": "Website/Website",
+    "website": "https://github.com/OCA/event",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": ["website_event_sale"],
+    "data": ["views/event_ticket.xml", "templates/event_ticket.xml"],
+}
--- a/website_event_sale_hide_ticket/models/event_event.py
+++ b/website_event_sale_hide_ticket/models/event_event.py
@@ -1,0 +1,21 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+
+from odoo import api, models
+
+
+class Event(models.Model):
+    _inherit = "event.event"
+
+    @api.onchange("event_type_id")
+    def _onchange_type(self):
+        super()._onchange_type()
+
+        if self.event_type_id.use_ticketing:
+            for type_ticket in self.event_type_id.event_ticket_ids:
+                ticket = self.event_ticket_ids.filtered(
+                    lambda t: t.product_id == type_ticket.product_id
+                    and t.price == type_ticket.price
+                )
+                if ticket:
+                    ticket.show_in_website = type_ticket.show_in_website
--- a/website_event_sale_hide_ticket/models/event_ticket.py
+++ b/website_event_sale_hide_ticket/models/event_ticket.py
@@ -1,0 +1,10 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+
+from odoo import fields, models
+
+
+class EventEventTicket(models.Model):
+    _inherit = "event.event.ticket"
+
+    show_in_website = fields.Boolean(default=True)
--- a/website_event_sale_hide_ticket/templates/event_ticket.xml
+++ b/website_event_sale_hide_ticket/templates/event_ticket.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <template
+        id="registration_template"
+        inherit_id="website_event_sale.registration_template"
+    >
+        <xpath expr="//t[@t-set='tickets']" position="attributes">
+            <attribute
+                name="t-value"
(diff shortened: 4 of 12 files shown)


website_event_sale_hide_ticket/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2020 Camptocamp SA
```
ajaniszewska-dev:
```suggestion
# Copyright 2021 Camptocamp SA
```

ajaniszewska-dev (approved):

website_event_sale_hide_ticket/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+{
+    "name": "Website Event Sale Hide Ticket",
+    "summary": "Allow to hide event ticket from the website",
+    "version": "13.0.1.0.0",
+    "development_status": "Alpha",
```
jcoux: Still in alpha?

website_event_sale_hide_ticket/models/event_event.py
```diff
@@ -0,0 +1,20 @@
...
+
+from odoo import api, models
+
+
+class Event(models.Model):
+    _inherit = "event.event"
+
+    @api.onchange("event_type_id")
+    def _onchange_type(self):
+        super(Event, self)._onchange_type()
```
jcoux:
```suggestion
        super()._onchange_type()
```

website_event_sale_hide_ticket/models/event_event.py
```diff
@@ -0,0 +1,20 @@
...
+
+        if self.event_type_id.use_ticketing:
+            for type_ticket in self.event_type_id.event_ticket_ids:
+                ticket = self.event_ticket_ids.filtered(
```
jcoux, line 16: Here you directly access to `event_ticket_ids`, but in the previous line, you did: `self.event_type_id.event_ticket_ids`.
Is it normal?

damdam-s: yes, event_type has tickets and event event too. Goal here is to sync a value from event_type to event_event

website_event_sale_hide_ticket/__manifest__.py
damdam-s: changed to beta but not stable as not used in production yet

website_event_sale_hide_ticket/models/event_event.py
```diff
@@ -0,0 +1,20 @@
...
+    def _onchange_type(self):
+        super()._onchange_type()
+
+        if self.event_type_id.use_ticketing:
+            for type_ticket in self.event_type_id.event_ticket_ids:
+                ticket = self.event_ticket_ids.filtered(
+                    lambda t: t.product_id == type_ticket.product_id
+                    and t.price == type_ticket.price
+                )
+                ticket.show_in_website = type_ticket.show_in_website
```
jcoux:
```suggestion
                if ticket:
                    ticket.show_in_website = type_ticket.show_in_website
```

jcoux (approved):

website_event_sale_hide_ticket/models/event_ticket.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+
+from odoo import fields, models
+
+
+class EventEventTicket(models.Model):
+    _inherit = "event.event.ticket"
+
+    show_in_website = fields.Boolean(default=True)
```
ivantodorovich, line 10: I think `website_published` is more consistent with odoo naming, as you can find this field in many website related models already

website_event_sale_hide_ticket/models/event_event.py
```diff
@@ -0,0 +1,21 @@
...
+    @api.onchange("event_type_id")
+    def _onchange_type(self):
+        super()._onchange_type()
+
+        if self.event_type_id.use_ticketing:
+            for type_ticket in self.event_type_id.event_ticket_ids:
+                ticket = self.event_ticket_ids.filtered(
+                    lambda t: t.product_id == type_ticket.product_id
+                    and t.price == type_ticket.price
+                )
```
ivantodorovich, line 19: Considering `super` "resets" the tickets from the event type, maybe it's possible to do something like this instead:

```py
def _onchange_type(self):
    res = super()._onchange_type()
    if self.event_type_id.use_ticketing:
        for ticket, type_ticket in zip(self.event_ticket_ids, self.event_type_id.event_ticket_ids):
            ticket.show_in_website = type_ticket.show_in_website
```

It'd avoid the product+price filtering that may not be accurate (nothing prevents the user to have two tickets with same price and product, only different seats limit / description / etc)


Unit tests would be good too :)

website_event_sale_hide_ticket/templates/event_ticket.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <template
+        id="registration_template"
+        inherit_id="website_event_sale.registration_template"
+    >
+        <xpath expr="//t[@t-set='tickets']" position="attributes">
+            <attribute
+                name="t-value"
+            >event.event_ticket_ids.filtered(lambda t: not t.is_expired and t.show_in_website)</attribute>
```
ivantodorovich, line 10: Isn't it possible to do this with `ir.rules` instead (the same way it's done for events) ?

See: https://github.com/odoo/odoo/blob/13.0/addons/website_event/security/event_security.xml

The downside of doing it with `ir.rules` is that the Odoo users will still be able to see these tickets on the website, which may be confusing, so it has to go with a qweb template that shows admins that the ticket is not published. 

See: https://github.com/odoo/odoo/blob/72cd8d0769097b21bbf94a34496888617058cb1b/addons/website_event/views/event_templates.xml#L199-L201 for example 

But in general it's a more solid approach IMO

ivantodorovich (approved):
Non blocking comments 👍🏻

PR #240, opened 2021-11-09
[13.0][MIG] crm_event: Migration to 13.0

yajo:
- Upstream merged both lead and opportunity form views.

@Tecnativa TT31318


pedrobaeza (approved):

victoralmau (approved):
Code review OK.

pedrobaeza:
Can you please squash together these commits?

Selección_005

And then you can proceed to merge (or tell me for me doing it).

crm_event/reports/event_type_report.py
```diff
@@ -94,10 +94,10 @@ def _groupby(self, clauses=()):
         result = sql.Composed(map(sql.SQL, clauses)).join(", ")
         return result
 
-    @api.model_cr
     def init(self):
         """(Re-)create report view."""
         tools.drop_view_if_exists(self.env.cr, self._table)
+        # pylint: disable=sql-injection
```
rousseldenis, line 100: @Yajo Why disabling this as it was correct? You should pass query params to the second argument of execute().

yajo: In this case I can't do it through params, because that would escape the SQL code. I need to execute the raw SQL code instead, as this is meant to create a SQL view. So, this is a false positive problem.

However thanks for raising the question, it reflects you reviewed thoroughly.

PR #242, opened 2021-11-09
[ADD] Event Registration Grade Module

rcordier77:


cvinh:
@rcordier77 please explain to the maintainers what your module does ?

rcordier77:
The module was created in order to allow to create event as Exam event and participant can be graded

PR #246, opened 2021-11-26
[12.0][FIX] website_event_crm: Take into account multiwebsite environments

sergio-teruel:
cc @Tecnativa TT33131

ping @chienandalu @CarlosRoca13

### DIFF
--- a/website_event_crm/README.rst
+++ b/website_event_crm/README.rst
@@ -117,7 +117,11 @@ Authors
 Contributors
 ~~~~~~~~~~~~
 
-* Jairo Llopis <jairo.llopis@tecnativa.com> (https://www.tecnativa.com/)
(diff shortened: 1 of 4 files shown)


chienandalu (approved):

CarlosRoca13 (approved):

pedrobaeza:
I think you should do `env["website"].get_current_website()` instead

sergio-teruel:
the get_current_website method returns the website from the domain.
In this case we send a email to a crm lead object in backend with a link to display all events from one category, in this link we need set the correct website and we only can obtain this from crm.lead company website.

pedrobaeza:
I think checking first `website_id` context key, and if not present, putting your code, can be good enough then.

pedrobaeza (approved):

PR #253, opened 2022-01-24
[13.0][MIG] event_track_location_overlap: Migration to 13.0

victoralmau:
Migration to 13.0

Please @ernestotejeda and @pedrobaeza can you review it?

@Tecnativa TT27583


pedrobaeza (commented):
Let's avoid the use of oca.decorators library, as it's very marginal. Remove the decorator `@foreach()`, and you call such method `_check_overlappable_one`, and let the other with the constraint for call record per record the other one.

event_track_location_overlap/models/event_track_location.py
```diff
@@ -14,15 +12,14 @@ class EventTrackLocation(models.Model):
     overlappable = fields.Boolean(help="Can this location have simultaneous tracks?")
 
     @api.constrains("overlappable")
```
pedrobaeza, line 14: You need also to split this for multi record support

victoralmau: Ok, if I have understood it correctly, the last change would be correct?

pedrobaeza: Yes.

pedrobaeza (approved):

ernestotejeda (approved):

PR #254, opened 2022-01-25
[13.0][MIG] event_sale_reservation: Migration to 13.0

victoralmau:
Migration to 13.0

Please @chienandalu and @pedrobaeza can you review it?

@Tecnativa TT31321


.travis.yml
```diff
@@ -27,10 +27,10 @@ jobs:
       after_success:
     - stage: test
       env:
-        - TESTS="1" ODOO_REPO="odoo/odoo" MAKEPOT="1"
+        - TESTS="1" ODOO_REPO="odoo/odoo" EXCLUDE="event_sale_reservation" MAKEPOT="1"
```
pedrobaeza: Why are you excluding the module from tests?

victoralmau: I have changed the file again to add the addon to the tests.
This addon is incompatible with: `event_registration_multi_qty` + `event_sale_registration_multi_qty` and for that reason it is excluded from Travis, and then only this addon is tested.

Related to: https://github.com/OCA/event/commit/10ea6248ad30e1286dde06a887d53da57b45a539

pedrobaeza: Well, you have to make them compatible. Start putting the problem, and seeing how to compatibilize them. Usually, this is done nullifying the modified behavior if in test mode and not detecting a specific context that is passed in the tests.

victoralmau: The comment that appears to exclude the addons is: 

> These modules makes totals wrong because they depend currently on count and not sum of qtys; integrating with them would require a glue module

It seems that it is not only related to the tests, I try to solve the compatibilities anyway?

pedrobaeza: Give it a try to see why the comment.

event_sale_reservation/wizards/registration_editor.py
```diff
@@ -47,7 +48,11 @@ def action_convert_to_registration(self):
             "actions": [
                 {"type": "ir.actions.act_window_close"},
                 {
-                    "context": dict(self.env.context, registering_reservations=False),
+                    "context": dict(
+                        self.env.context,
+                        registering_reservations=False,
+                        skip_event_sale_registration_multi_qty=True,
```
chienandalu, line 57: If we go this way, give at least an explanation of why this context key is needed. Anyway. This doesn't make both modules compatible, so maybe a note would be needed in the ROADMAP as well

chienandalu (commented):
Tested functionally. As for the `event_registration_multi_qty`:

victoralmau:
Changes done. Any other comments to take into account to apply it?

chienandalu (approved):
Thanks :)

pedrobaeza (approved):

PR #262, opened 2022-03-07
[13.0][FIX] website_event_filter_city: search input

chienandalu:
- Refactor to try to reuse as much from the core as possible. Anyway, as there's no external method to compose the domain we still have to redo some things in order to add our city filter stuff.
- Consider the new name search term.
- Fix test: it didn't expect online events when filtering by country, but Odoo core allows it explicitly.

cc @Tecnativa TT34958

### DIFF
--- a/website_event_filter_city/controllers/main.py
+++ b/website_event_filter_city/controllers/main.py
@@ -1,47 +1,66 @@
 # Copyright 2016-2017 Jairo Llopis <jairo.llopis@tecnativa.com>
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
-from odoo import _, http
+from odoo import _
+from odoo.http import request, route
 
 from odoo.addons.website_event.controllers.main import WebsiteEventController
 
 
 class WebsiteEvent(WebsiteEventController):
-    @http.route()
+    @route()
     def events(self, page=1, **searches):
-        # Get current render values
-        result = super().events(page, **searches)
-        values = result.qcontext
-        searches = values["searches"]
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Isn't possible to reuse upstream code calling super and avoid future similar compatibility problems?

chienandalu:
I'd say main problem is that the core method is encaupsulated and the counter values need to be recalculated with the new search criteria. @victoralmau was involved in the migration. Surely he could tell better

victoralmau:
> 

Exactly, without a `_prepare()` function (or similar) in `website_event` it is not possible to call super and reduce code. It is necessary, for example, to re-do the code for the total number of records or for the domain that we must apply.

victoralmau (approved):

rafaelbn (approved):

pedrobaeza:
@chienandalu please try again with the technique we talked about of injecting the context and then overwriting search method.

chienandalu:
> @chienandalu please try again with the technique we talked about of injecting the context and then overwriting search method.

Done. Anyway, surely not as elegant and minimalist as you expected :sweat_smile: These complex controller methods tend have encapsulated behaviors that force us to redo some things anyway.

pedrobaeza:
Indeed is a bit ugly, but I hope we reduce the friction in other upstream patches. I will update CI for using GH actions and merge. Is this needed in v14?

chienandalu:
> Is this needed in v14?

The module isn't yet migrated

victoralmau (approved):

rafaelbn (approved):

PR #263, opened 2022-03-16
[MIG] event_session: Migration to 14.0

chienandalu:
Migration highlights:

- `event` has changed the `state` field for a relation to configurable stages. We had to adapt to this new workflow.
- Some old features were dropped as obsolete, so we drop them from event sessions as well (`seats_min`).
- `event_event.seats_availability` select field changed to the new `seats_limited` boolean one. We adopt the same change in event sessions and a migration scrip is added to convert the data.
- Record rules are updated to adapt to the new multicompany behavior.
- Some fixes in tests.

cc @Tecnativa TT34351

please review @ernestotejeda @pedrobaeza


event_session/models/event_mail.py
```diff
@@ -50,30 +50,21 @@ def _compute_done(self):
...
                 )
 
-    @api.depends(
-        "event_id.state",
-        "event_id.date_begin",
-        "interval_type",
-        "interval_unit",
-        "interval_nbr",
-    )
+    @api.depends()
```
pedrobaeza: Why removing dependencies here?

pedrobaeza (approved):

event_session/models/event_mail.py
ivantodorovich: Same dependencies than super I think.
But I do think these are missing:

- `session_id.date_begin`
- `session_id.date_end`

ivantodorovich (approved):

event_session/models/event_mail.py
chienandalu: Thanks @ivantodorovich you're right :+1:

pedrobaeza:
Travis is now red

pedrobaeza:
It seems the depends you have re-added are not OK

event_session/models/event_mail.py
```diff
@@ -0,0 +1,95 @@
...
+                    else len(event_mail.mail_registration_ids)
+                    == len(event_mail.session_id.registration_ids)
+                    and all(line.mail_sent for line in event_mail.mail_registration_ids)
+                )
+
+    @api.depends(
+        "event_id.state",
+        "event_id.date_begin",
+        "session_id.date_begin",
+        "session_id.date_end",
```
ivantodorovich: You can keep only these two.
`api.depends` will add new ones when overriding, but it will keep the original ones

event_session/migrations/14.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,24 @@
...
+from openupgradelib import openupgrade
+
+
+def rename_event_session_seats(env):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        ALTER TABLE event_session
+        ADD COLUMN seats_limited boolean""",
+    )
```
ernestotejeda, line 12: Why not to use here `add_fields` method of openupgradelib

event_session/models/event.py
```diff
@@ -39,6 +39,9 @@ class EventEvent(models.Model):
         string=" # No of Confirmed Registrations",
         store=True,
     )
+    generate_sessions_button_visible = fields.Boolean(
+        compute="_compute_generate_sessions_button_visible",
+    )
```
ernestotejeda: What about to have a relate field (`stage_id.pipe_end`) instead?

event_session/migrations/14.0.1.0.0/pre-migration.py
pedrobaeza: It's not needed anymore since v13, as Odoo does all the needed stuff.

event_session/models/event_mail.py
```diff
@@ -0,0 +1,95 @@
...
+                    True
+                    if event_mail.event_id.sessions_count > 0
+                    and not event_mail.session_id
+                    else len(event_mail.mail_registration_ids)
+                    == len(event_mail.session_id.registration_ids)
+                    and all(line.mail_sent for line in event_mail.mail_registration_ids)
+                )
+
+    @api.depends(
+        "event_id.state",
```
ernestotejeda:
```suggestion
        "event_id.stage_id",
```
That's why travis is red

event_session/models/event.py
chienandalu: Ok, I like it :)

event_session/models/event_mail.py
chienandalu: Weird, that's wasn't on my local code. I'm pushing again anyway

event_session/models/event.py
chienandalu: I like it

event_session/models/event_mail.py
chienandalu: Thanks :) I'm never completly sure about that

chienandalu:
@ernestotejeda can you update your review?

ernestotejeda (approved):

PR #265, opened 2022-03-17
[14.0][MIG] event_registration_multi_qty: Migration to 14.0

chienandalu:
Standard migration

cc @Tecnativa TT34348

please review @sergio-teruel @ernestotejeda

PR #267, opened 2022-03-24
[MIG] event_session_registration_multi_qty: Migration to 14.0

chienandalu:
Standard migration.

Depending on:

- [x] #263 
- [x] #265 

cc @Tecnativa TT34352

please review @ernestotejeda @pedrobaeza


pedrobaeza:
/ocabot migration event_session_registration_multi_qty

Please rebase

chienandalu:
Rebased

ernestotejeda (approved):

event_session_registration_multi_qty/models/event_session.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class EventSession(models.Model):
+    _inherit = "event.session"
+
+    @api.depends("registration_ids.qty")
+    def _compute_seats(self):
+        for session in self:
+            if not session.event_id.registration_multi_qty:
+                return super()._compute_seats()
```
ivantodorovich, line 13:
```suggestion
        records = self.filtered("event_id.registration_multi_qty")
        super(EventSession, self - records)._compute_seats()
        for session in records:
```

Better to call super outside of the loop, otherwise you loose all the recordset optimizations

event_session_registration_multi_qty/models/event_session.py
```diff
@@ -0,0 +1,50 @@
...
+                "seats_unconfirmed": 0,
+                "seats_reserved": 0,
+                "seats_used": 0,
+                "seats_available": 0,
+                "seats_available_expected": 0,
+                "seats_available_pc": 0,
+            }
+            registrations = self.env["event.registration"].read_group(
+                [
+                    ("session_id", "=", session.id),
```
ivantodorovich, line 24: Better to do this from outside the loop, to reduce the number of queries to compute this thing.
Otherwise you run a query per record, that's not efficient.

You probably need to use raw sql to achieve it, like it's done in the original method

Somthing like this perhaps

```py
    state_field = {
        'draft': 'seats_unconfirmed',
        'open': 'seats_reserved',
        'done': 'seats_used',
    }
    if self.ids:
        query = """
            SELECT session_id, state, sum(qty)
             FROM event_registration
             WHERE session_id IN %s
             AND state IN ('draft', 'open', 'done')
             GROUP BY session_id, state
        """
        self.env['event.registration'].flush(['session_id', 'state', 'qty'])
        self.env.cr.execute(query, (tuple(self.ids),))
        res = self.env.cr.fetchall()
        for session_id, state, qty in res:
            results[session_id][state_field[state]] = num
```

rafaelbn (approved):

event_session_registration_multi_qty/models/event_session.py
chienandalu: Thanks @ivantodorovich I'll try to get some time and check your comments :)

pedrobaeza:
Oh, sorry, I didn't see the last comment. The improvement can go in a new PR though.

PR #269, opened 2022-04-21
[FIX] website_event_crm: another company's event

chienandalu:
The company field isn't mandatory for event types, so we have to check that a lead auto invites to events available in the company of the lead.

- [x] Ensure sending company wise invitations to events.
- [x] Ensure right base_url for the sent link
- [x] Disengaged `crm_event` test data declaration for easy  reuse
- [x] Regression tests

cc @Tecnativa TT36027

### DIFF
--- a/website_event_crm/models/crm_lead.py
+++ b/website_event_crm/models/crm_lead.py
@@ -13,7 +13,8 @@ _logger = logging.getLogger(__name__)
 class CRMLead(models.Model):
     _inherit = "crm.lead"
 
-    event_type_website_url = fields.Char(compute="_compute_event_type_url",)
+    event_type_website_url = fields.Char(compute="_compute_event_type_url")
+    auto_invite_warning = fields.Boolean(compute="_compute_auto_invite_warning")
 
     @api.depends("event_type_id")
     def _compute_event_type_url(self):
@@ -27,9 +28,35 @@ class CRMLead(models.Model):
             if events:
                 lead.event_type_website_url = "/event?type=%d" % lead.event_type_id.id
 
+    @api.depends("event_type_id", "company_id")
+    def _compute_auto_invite_warning(self):
+        """Used for warning purposes in the lead/opportunity form view"""
+        self.auto_invite_warning = False
+        for lead in self.filtered("event_type_website_url"):
+            # If there's no website for the lead/opportunity company we'll warn the user
+            # that the invitation won't be sent.
+            website = lead.company_id and self.env["website"].search(
+                [("company_id", "=", lead.company_id.id)]
+            )
+            if not lead.company_id or not website:
+                lead.auto_invite_warning = True
+
     @api.model
     def _cron_auto_invite_website_event_type(self):
-        """Invite automatically leads/opportunities to website event categories."""
+        """Invite automatically leads/opportunities to website event categories. We
+        should be aware of multi-company + multi-website scenarios in which we want
+        the link sent to the customer launched only when there are events for their
+        company leads. Moreover, we want to send the proper base url link for the
+        proper filtered events.
+        - crm.leads company isn't mandatory (although unusual). We ignore these cases
+          as we can't determine properly what to send to the user.
+        - event.event website isn't mandatory: This means that the event will be
+          available on all the sites belonging to the company.
+        - There could multiple websites in one company: if the event has the link to
+          one of them, we'll get the proper base url.
+        - There could be no website in a company. In this case, it won't make sense
+          to send anything to the customer as he won't be able to reach it.
+        """
         # Find opportunities
         leads = self.search(
             [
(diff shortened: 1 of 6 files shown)


website_event_crm/models/crm_lead.py
```diff
@@ -65,6 +48,30 @@ def _cron_auto_invite_website_event_type(self):
...
+                ("event_type_id", "!=", False),
+                ("stage_id.auto_invite_website_event_type", "=", True),
+            ]
+        )
+        for company in [x.company_id for x in leads]:
+            company_leads = leads.filtered(lambda l: l.company_id == company)
+            available_event_types = company_leads.mapped("event_type_id")
+            published_events = (
+                self.env["event.event"]
+                .search(available_event_types._published_events_domain())
```
pedrobaeza: Do the `search` only once outside the loop and filter inside the loop. At the end, it's the same to look for all events of all types, and then filter events by company, than filter types by company, and filter again events by company.

pedrobaeza (changes requested):
Don't you need to pass `force_company` or similar to the method that makes the invitations? Or how the context is managed here? I think a comment about why the loop is also advisable.

website_event_crm/models/crm_lead.py
```diff
@@ -65,6 +48,30 @@ def _cron_auto_invite_website_event_type(self):
...
+        """Invite automatically leads/opportunities to website event categories."""
+        # Find opportunities
+        leads = self.search(
+            [
+                ("email_from", "!=", False),
+                ("event_type_id", "!=", False),
+                ("stage_id.auto_invite_website_event_type", "=", True),
+            ]
+        )
+        for company in [x.company_id for x in leads]:
```
yajo: I guess that if you do it like this, you'll repeat companies. Try with:

```suggestion
        for company in leads.mapped("company_id"):
```

website_event_crm/models/crm_lead.py
```diff
@@ -65,6 +48,30 @@ def _cron_auto_invite_website_event_type(self):
...
+            company_leads = leads.filtered(lambda l: l.company_id == company)
+            available_event_types = company_leads.mapped("event_type_id")
+            published_events = (
+                self.env["event.event"]
+                .search(available_event_types._published_events_domain())
+                .filtered(lambda e: e.company_id == company or not e.company_id)
+            )
+            published_event_types = published_events.mapped("event_type_id")
+            if not published_event_types:
+                return
```
yajo: You probably need to change this too:

```suggestion
                continue
```

yajo (changes requested):

website_event_crm/models/crm_lead.py
chienandalu: Sure! :sweat_smile:

website_event_crm/models/crm_lead.py
chienandalu: I was considering leads with falsey `company_id` (the proper way would be `{x.company_id for x in leads}` but thinking twice, for this cases maybe we should avoid sending the invitation. And thinking again as I write this, maybe what I should consider in first place is that the company has a website where the user can view the notified events...

website_event_crm/models/crm_lead.py
chienandalu: Done

pedrobaeza (commented):
Still missing the explanation in the code about the loop for not forgetting why we do this in the future.

chienandalu:
Added multiwebsite base url support. Not an easy choice, but finally I decided to ignore those published events with no website set:

- An event could be multicompany and multiwebsite and be available for every website and every company
- An event could be multicompany and multiwebsite, so it will be available across al the company's websites
- A company could have several websites, what would the right link be here???
- A lead could have no company nor website (ok, we discard these...)

For now, I've gone for filtering events by their website, but I feel it's not right. What do you think?

website_event_crm/__manifest__.py
```diff
@@ -11,7 +11,7 @@
     "maintainers": ["Yajo"],
     "license": "AGPL-3",
     "installable": True,
-    "depends": ["crm_event", "website_event"],
+    "depends": ["crm_event", "website_event", "sale_stock"],
```
pedrobaeza: Why this new dependency?

website_event_crm/models/crm_lead.py
```diff
@@ -65,6 +48,47 @@ def _cron_auto_invite_website_event_type(self):
             except Exception:
                 _logger.exception("Failure trying to invite to website event type.")
 
+    @api.model
+    def _cron_auto_invite_website_event_type(self):
```
pedrobaeza, line 45: Can we avoid to reorder this for reducing the diff?

website_event_crm/models/crm_lead.py
```diff
@@ -65,6 +48,47 @@ def _cron_auto_invite_website_event_type(self):
...
+                ("email_from", "!=", False),
+                ("event_type_id", "!=", False),
+                ("stage_id.auto_invite_website_event_type", "=", True),
+            ]
+        )
+        available_event_types = leads.mapped("event_type_id")
+        published_events = self.env["event.event"].search(
+            available_event_types._published_events_domain()
+        )
+        # Leads with falsey company won't get invitations
```
pedrobaeza:
```suggestion
        # Leads with falsy company won't get invitations
```

Shouldn't this raise an exception if we include any of the ignored ones? If not, people will continue thinking that they have been invited, and wonder why it didn't get. Anyways, I thought that the final criteria is when no website is set.

pedrobaeza (commented):
Please rebase, as I have fixed current CI and switch to GH actions

website_event_crm/__manifest__.py
chienandalu: Sorry, I had it for local testing purposes

website_event_crm/models/crm_lead.py
chienandalu: Well, if the cron is the one inviting I don't think so.

> I thought that the final criteria is when no website is set.

But in the event... anyway, as I put in my previous comment, the main problem here is that website isn't mandatory on events.

The issues are two:

- Avoid sending an invitation when no events are available in the category of the event.
- Send the right base_url for each case.

A quick look over the possible values:

|  *required*  | crm.lead | event.event |
|------------|----------|-------------|
| company_id | False    | True        |
| website_id | False    | False       |

|  *default*   | crm.lead | event.event |
|------------|----------|-------------|
| company_id | True     | True        |
| website_id | False    | False       |

We can take for granted that the leads will have the company set. What I think I'll do is to send the notifications by website but also send those from events with no website with the base url of the first website in the company. So in every case the user will be able to check the event list and no invitation will be left.

pedrobaeza: Then put a warning (in the crm.lead I'm interpreting) in the top like other places when a condition for not being sent is met, saying that no mail will be sent due to ...

website_event_crm/models/crm_lead.py
chienandalu: Done

pedrobaeza (approved):

victoralmau (approved):

pedrobaeza:
@yajo please update your review

yajo (approved):

PR #272, opened 2022-04-27
[13.0][IMP] crm_event: Promote module to stable

pedrobaeza:
They comply with the requirements and are needed by manifestoo.

@Tecnativa


pedrobaeza:
Working:

/ocabot merge patch

PR #275, opened 2022-05-20
[15.0] [MIG+REF / RFC] `event_session`: Migration to 15.0

ivantodorovich:
We've done a huge refactoring to this module in order to support core features found in regular events, but also for sessions.

Some highlights:

* Dropped the `event_mail` dependency, make it work with core mail schedulers instead.
* Support core voucher / badge reports for sessions.
* Now `event.session` inherits from `event.event` (like `product.product` and `product.template`)
* Make it easier to add more sessions to an event (refactored wizard, using recurrency-like rules)
* Play nice with regular events (events that don't use sessions) -- it's possible to have both kinds at the same time
* Implement (or attempt to) all core `event` features with sessions, but nothing more.

(and naturally there's also the [`event_sale_session`](https://github.com/OCA/event/pull/276) module, and some others to implement them in website and pos)

@chienandalu I understand you're using this module in previous versions, and such a big change may not be welcomed. Consider this a RFC rather than a migration PR, to see if we can join forces


event_session/models/event_session.py
```diff
@@ -0,0 +1,524 @@
...
+
+    def _sync_session_update(self, vals):
+        """Handles write on multiple sessions at once from the UX"""
+        update = vals.pop("session_update", "this")
+        if update not in ("subsequent", "all"):
+            return
+        if len(self) > 1:
+            raise ValidationError(
+                _("You cannot use session_update when writing on recordsets")
+            )
```
petrus-v, line 496: I'm just wondering why we are not using `self.ensure_one()` here ?

```suggestion
        self.ensure_one()
```

event_session/models/event_session_timeslot.py
```diff
@@ -0,0 +1,43 @@
...
+        try:
+            hours, minutes = [int(part) for part in name.split(":")]
+            if not (0 <= hours <= 23):
+                raise Exception()
+            if not (0 <= minutes <= 60):
+                raise Exception()
+        except Exception:
+            raise ValidationError(
+                _("The timeslot has to be defined in HH:MM format")
+            ) from None
```
petrus-v: I think we could use time library to assert format a get a nicer report getting whole exception history if we want to raise ValidationError instead ValueError:

```suggestion
        import time
        [...]
        try:
            struct_time = time.strptime(name, '%H:%M')
            hours, minutes = struct_time.tm_hour, struct_time.tm_min
        except ValueError ex:
            raise ValidationError(
                _("The timeslot has to be defined in HH:MM format")
            ) from ex
            
```

petrus-v (approved):
LGTM

few styling remarks

event_session/models/event_session.py
ivantodorovich: Well, ensure_one will fail if len == 0, which may happen sometimes due to programatically setting field values on empty recordset (that shouldn't fail, just shouldn't do anything)

But more importantly, I like to raise a clear error message

event_session/models/event_session_timeslot.py
ivantodorovich: Thanks for the tip! I like it indeed :)
Code updated

pedrobaeza:
@chienandalu can you please review this?

gregory-moka (approved):

pedrobaeza:
@chienandalu is it OK for you this refactoring?

chienandalu:
Looks good (thanks for the work @ivantodorovich !), but I have to do an in depth review

ivantodorovich:
Thanks a lot @chienandalu :)

pedrobaeza:
Should we merge then, @chienandalu, or are you going to do the review before the merge?

PR #280, opened 2022-08-09
[14.0][FIX] event_registration_partner_unique: improve README

LudLaf:
After being merged, a clearer explanation has been produced about how this module works. 

@Tecnativa TT33815

### DIFF
--- a/event_registration_partner_unique/readme/USAGE.rst
+++ b/event_registration_partner_unique/readme/USAGE.rst
@@ -1,8 +1,7 @@
 #. Go to **Events** and choose an event or create a new one.
 #. Enable **Forbid Duplicates**.
 #. Go to **Attendees**.
-#. Create a new attendee, starting with **Booked by**.
-#. Several attendees can have the same *Booked by* person or company, but if
-   you try to register an attendee by **Attendee Name** or by
-   **Attendee Partner** more than once for the same event, the system won't
-   allow it and will show a *Validation Error*.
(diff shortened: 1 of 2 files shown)


event_registration_partner_unique/readme/USAGE.rst
```diff
@@ -1,8 +1,6 @@
 #. Go to **Events** and choose an event or create a new one.
 #. Enable **Forbid Duplicates**.
 #. Go to **Attendees**.
-#. Create a new attendee, starting with **Booked by**.
-#. Several attendees can have the same *Booked by* person or company, but if
-   you try to register an attendee by **Attendee Name** or by
-   **Attendee Partner** more than once for the same event, the system won't
-   allow it and will show a *Validation Error*.
+#. Create a new attendee.
+#. If you try to register an attendee more than once for the same event, the
```
pedrobaeza, line 5: No indication here about which field should be filled?

event_registration_partner_unique/readme/USAGE.rst
```diff
@@ -1,8 +1,6 @@
...
+   system won't allow it.
```
pedrobaeza:
```suggestion
#. If you try to fill successive attendees with the same contact filled in
   the "Attendee Partner" field, the system won't allow it. That can
   happen for example if the same email is used several times.
```

pedrobaeza (approved):

PR #281, opened 2022-08-10
[15.0] [MIG] `website_event_sale_hide_ticket`: Migration to 15.0  (renamed to `website_event_ticket_published`) 

Camille0907:
Migration of `website_event_sale_hide_ticket` (renamed to `website_event_ticket_published`) module from v13 to v15


ivantodorovich (approved):

ivantodorovich:
In `15.0`, the `event.ticket` model has been moved from `event_sale` module to the `event` module.

This means the dependency could be changed to `website_event` (instead of `website_event_sale`), to support this feature even if `event_sale` is not installed. The module could then be renamed to `website_event_hide_ticket` (or maybe `website_event_ticket_published` for consistency with the website publish/unpublish concept in odoo).

Just a suggestion, not blocking

chienandalu (approved):

PR #213, opened 2021-05-11
[13.0][mig] website_event_snippet_calendar

damdam-s:
superseedes #191

PR #288, opened 2022-12-20
[15.0] [MIG] `website_event_snippet_calendar`: Migration to 15.0

Camille0907:
Migration of `website_event_snippet_calendar` module from v13  (from [this pr](https://github.com/OCA/event/pull/213)) to v15

I tried to take into accounts the last comments from the v13 pr, more recent than the latest fore-push on the branch : 
- https://github.com/OCA/event/pull/213#issuecomment-859693628: it's already the case
- https://github.com/OCA/event/pull/213#discussion_r651512054: the treatment of the url arguments is moved in the JS code, which send a dict to the python code
- https://github.com/OCA/event/pull/213#discussion_r651512741: I kept a generic name, so that we could use other arguments in the future, for exemple to filter on something else than event_type

I also made some improvements: 
- The filter is taken into account for enabling the right days on the calendar
- The option in the website has been a little bit changed:
    - The buttons to change the number of events have been renamed
    - The number of events can't go under 1 (0 was authorized, but not managed, before)
- There is now a message if the list is empty, to explain that no event is available


damdam-s (approved):

website_event_snippet_calendar/controllers/main.py
```diff
@@ -39,13 +42,24 @@ def days_with_events(self, start, end):
...
 
+    def get_event_type_from_parameters(self, parameters=None):
+        if not parameters:
+            return None
+        event_type = parameters.get("type", None)
+        if event_type:
+            try:
+                event_type = int(event_type) or None  # 0 is not a valid ID
+            except ValueError:
+                event_type = None
```
yankinmax: @Camille0907 I don't understand what you try to achieve here:
```
>>> int(0)
0
>>> int('0')
0
```
can you pls clarify?

yankinmax: Maybe you can give some examples of `event_type`?

yankinmax (approved):

website_event_snippet_calendar/controllers/main.py
Camille0907: `event_type` is supposed to be the id of a `event.type` (<=> an "Event Template") existing in the system. Events can be linked to an event template, so we want here to have the possibility to filter on the events linked to a specific "Event template" : in this case the ID of the event template has to be stored in this variable `event_type`.
This variable comes from the URL of the page, from a GET parameter : `https://my-site/specific-path?type=5`. Thus we can't be sure to have a valid ID : it can be something else than a number (caught by the `except`), or it can be 0 (IDs begin from 1) : since 0 is not valid we want the same behavior as in the `except`.
So : 
- if event_type is not a number -> we go in the `except` -> None
- if event_type is 0 (or '0') -> int(event_type) = 0 <=> False -> we go in the second part of the `or` -> None
- if event_type is a number >= 1 -> int(event_type) != 0 <=> True -> we stay in the first part of the `or` -> event_type stores the valid id

I hope I'm clear and answering to your remark

yankinmax: Nice sanitizing of `event_type`, thanks for deep expalanation

Camille0907: Makes me realize that it's completely useless : when the function is called somewhere, it checks if the returned result is equivalent to True before going further, to avoid using `None` -> returning 0 have exactly the same impact than returning None, no need to treat it as a special case
I'll simplify, thanks !

Camille0907:
Added one commit to split the function that allows to get events to display, to ease the overriding of the domain
I don't know if we want it in the migration commit or not

chienandalu (approved):

PR #293, opened 2023-01-12
[15.0][MIG] partner_event: Migration to version 15.0

stefan-tecnativa:
cc @Tecnativa TT41018

@chienandalu @victoralmau please review


chienandalu (changes requested):
Check pre-commit please

chienandalu (changes requested):
Functional test ok.

- Fix your migration commit message as is reversed with its content
- Let's update that old icon
icon(1)
icon(1)

partner_event/tests/test_event_registration.py
```diff
@@ -75,7 +75,7 @@ def test_button_register(self):
         event_1 = self.event_0.copy()
         wizard = self.env["res.partner.register.event"].create({"event": event_1.id})
         active_ids = [self.partner_01.id, self.registration_02.attendee_partner_id.id]
-        wizard.with_context({"active_ids": active_ids}).button_register()
+        wizard.with_context(**{"active_ids": active_ids}).button_register()
```
chienandalu: It feels more natural to do it with key arguments instead of a dict unpack when the number is just one or two
```suggestion
        wizard.with_context(active_ids=active_ids).button_register()

Comment (stefan-tecnativa) [reply partner_event/tests/test_event_registration.py]: Thanks for the proposal, I will take it into account for the next one.

**Review by chienandalu — APPROVED** [review]

**Review by victoralmau — APPROVED** [review]

File: partner_event/__manifest__.py [hunk 2]
```diff
@@ -9,7 +9,7 @@
 
 {
     "name": "Link partner to events",
-    "version": "14.0.1.0.1",
+    "version": "15.0.1.0.1",
```
pedrobaeza:
```suggestion
   "version": "15.0.1.0.0",
```

pedrobaeza (changes requested):
Please squash together these translation commits:

Selección_025

and also include the SVG in partner_event/static/description folder

partner_event/__manifest__.py
stefan-tecnativa: I didn't realize it, thank you

stefan-tecnativa:
> Please squash together these translation commits:
> 
> Selección_025
> 
> and also include the SVG in partner_event/static/description folder

Would it be appropriate to do this with all translation commits as well?

pedrobaeza:
> Would it be appropriate to do this with all translation commits as well?

No, because you will lose the attribution to the corresponding translator. That's why I only say to merge together the translations from the same translator.

pedrobaeza:
On the new push you have made, I'm not seeing any of the Italian translation commits. Have you removed them or squash with another? I'm now also seeing 2 Swedish translation commits coming from the same person that can be squashed together.

stefan-tecnativa:
> > Would it be appropriate to do this with all translation commits as well?
> 
> No, because you will lose the attribution to the corresponding translator. That's why I only say to merge together the translations from the same translator.

Okay, understood. Thanks.

stefan-tecnativa:
> On the new push you have made, I'm not seeing any of the Italian translation commits. Have you removed them or squash with another? I'm now also seeing 2 Swedish translation commits coming from the same person that can be squashed together.

Change done.

pedrobaeza (approved):
/ocabot merge nobump
/ocabot migration partner_event

PR #295, opened 2023-01-27
[15.0][MIG] event_email_reminder: Migration to 15.0

victoralmau:
Migration to 15.0

Please @CarlosRoca13 and @pedrobaeza can you review it?

@Tecnativa TT41010


event_email_reminder/models/event.py
```diff
@@ -0,0 +1,73 @@
...
+
+from odoo import api, fields, models
+
+
+class EventEvent(models.Model):
+    _inherit = "event.event"
+
+    @api.model
+    def _send_event_template(self, events, template, partner_ids):
+        ctx = self.env.context.copy()
```
chienandalu: I wonder if this is really necessary. Wouldn't a regular dict be enough as `with_context` merges the updated keys with the current context already?

chienandalu (approved):
:+1:

chienandalu:
An icon upgrade would be nice :)

event_email_reminder/data/event_email_reminder_data.xml
```diff
@@ -1,52 +1,52 @@
...
+                </tr>
+        </table>
     </field>
     </record>
     <record id="ir_cron_event_email_reminder_action" model="ir.cron">
         <field name="name">Event Email Reminder</field>
         <field name="active" eval="True" />
         <field name="model_id" ref="event.model_event_event" />
-        <field name="code">model.run_event_email_reminder(7, False, False, None)</field>
+        <field name="code">model.run_event_email_reminder(7, False, None)</field>
```
pedrobaeza: You can separate this file by model: `ir_cron_data.xml` and `mail_template_data.xml`. About this cron, we are changing the behavior of the module: on previous version, "new" events were not included. Now, all the events are included. The change on the data model has been to have a flexible stage system, through a new model `event.stage`. I think you should include a flag in it to exclude or not from these notifications, and pre-mark as excluded the stage "New".

pedrobaeza (changes requested):

event_email_reminder/data/event_email_reminder_data.xml
victoralmau: Ok, changes done.

CarlosRoca13 (approved):

event_email_reminder/models/event_event.py
```diff
@@ -0,0 +1,77 @@
...
+        :param int template_id: id of a template or None.
+        :param list(int) partner_ids: list of IDs of the partners we
+          want to notify.
+        """
+        today = fields.Date.context_today(self)
+        limit_date = today + timedelta(days=days)
+        if draft_events:
+            domain = [("stage_id.exclude_from_email_reminder", "=", True)]
+        else:
+            domain = [("stage_id.exclude_from_email_reminder", "=", False)]
```
pedrobaeza:
```suggestion
            domain = []
```

pedrobaeza (approved):
/ocabot merge nobump

PR #296, opened 2023-01-27
[15.0][MIG] event_sale_reservation: Migration to 15.0

victoralmau:
Migration to 15.0

Please @pedrobaeza and  @chienandalu can you review it?

@Tecnativa TT41011


event_sale_reservation/tests/test_event_sale.py
```diff
@@ -99,7 +99,7 @@ def wizard_step_2(self, wizard1):
         # Get form from 2nd chained action
         action = multi_action["actions"][1]
         return Form(
-            self.env[action["res_model"]].with_context(action["context"]),
+            self.env[action["res_model"]].with_context(**action["context"]),
```
pedrobaeza, line 103: In this case it doesn't matter, as there's no previous context

event_sale_reservation/models/sale_order_line.py
```diff
@@ -28,5 +28,14 @@ class SaleOrderLine(models.Model):
     @api.depends("event_registration_ids")
     def _compute_event_registration_count(self):
         """Get count of related event registrations."""
+        items_data = self.env["event.registration"].read_group(
```
pedrobaeza: There's an ongoing debate about what is more performant lately, as the cache avoids to do constant queries to DB as the `read_group` does, but it also reads more data. If v17 contains the `read_group` option reading from cache, it will be clear. Meanwhile, I think you can leave it as it was for minimizing diff.

event_sale_reservation/views/product_template_view.xml
```diff
@@ -6,17 +6,18 @@
         <field name="name">Allow products that are event registrations</field>
         <field name="model">product.template</field>
         <field name="groups_id" eval="[(4, ref('event.group_event_user'))]" />
-        <field name="inherit_id" ref="event_sale.event_sale_product_template_form" />
+        <field name="inherit_id" ref="product.product_template_form_view" />
         <field name="arch" type="xml">
-            <field name="event_ok" position="attributes">
+            <field name="detailed_type" position="attributes">
```
pedrobaeza: Seeing this, for me it seems what you have to do is to add a new `detailed_type`: `event_reservation`, instead of all this stuff.

pedrobaeza (changes requested):

event_sale_reservation/tests/test_event_sale.py
victoralmau: This is only to avoid pre-commit error: "_Context overridden using dict. Better using kwargs..._"

victoralmau:
Changes done.

event_sale_reservation/readme/USAGE.rst
```diff
@@ -0,0 +1,58 @@
+To know how many reservations exist for a given event type:
+
+#. Go to *Events > Configuration > Event Categories* and pick or create one.
```
chienandalu:
```suggestion
#. Go to *Events > Configuration > Event Templates* and pick or create one.
```

event_sale_reservation/readme/USAGE.rst
```diff
@@ -0,0 +1,58 @@
...
+   that info from the sales order customer.
+#. After that's done, click on *Apply*.
+
+At this point, the sales order lines will be modified to include the ticket
+product instead of the reservation product, and the event reservations have
+been created, linked to those lines.
+
+If the ticket was free, the registrations are confirmed. Otherwise, they are
+kept as draft until an invoice is created for the sales order, and paid. But
+that is just upstream ``event_sale`` module in action.
```
chienandalu:
```suggestion
If the event is set to autoconfirmation, the registrations are confirmed., otherwise, they are
kept as draft until an invoice is created for the sales order, and paid. But
that is just upstream ``event_sale`` module in action.
```

chienandalu (approved):
Tested :+1: Just two readme suggestions:

event_sale_reservation/migrations/15.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,12 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.logged_query(
+        env.cr,
+        """UPDATE product_product SET detailed_type = 'event_reservation'
+        WHERE event_reservation_ok IS TRUE""",
```
pedrobaeza, line 11: Boolean, so no need to do an extra equal operation

```suggestion
        WHERE event_reservation_ok""",
```

pedrobaeza (approved):
/ocabot merge nobump

PR #290, opened 2023-01-02
[15.0][MIG]website_event_require_login: Migration to 15.0

bizzappdev:

PR #306, opened 2023-03-23
[15.0][MIG] website_event_require_login: Migration to version 15.0

stefan-tecnativa:
cc @Tecnativa TT42046

@chienandalu please review!


chienandalu:
This is being migrated here as well: https://github.com/OCA/event/pull/290

PR #312, opened 2023-05-10
[15.0][MIG] event_mail: Migration to version 15.0

ernestotejeda:
Cc @Tecnativa TT42036


event_mail/models/event_mail.py
```diff
@@ -28,14 +38,16 @@ def _default_scheduler_template_ids(self):
...
+                % self.env.ref("event.event_subscription").id,
             },
             {
                 "notification_type": "mail",
                 "interval_nbr": 10,
                 "interval_unit": "days",
                 "interval_type": "before_event",
-                "template_id": self.env.ref("event.event_reminder").id,
+                "template_ref": "mail.template, %i"
+                % self.env.ref("event.event_reminder").id,
```
chienandalu:
```suggestion
                "template_ref": f"mail.template, {self.env.ref("event.event_reminder").id}",

File: event_mail/tests/test_event_mail.py [hunk 2]
```diff
@@ -86,7 +87,7 @@ def test_event_template_no_config(self):
         event = self.env["event.event"].create(vals)
         self.assertEqual(
             len(event.event_mail_ids),
-            0,
+            len(event._default_event_mail_ids()),
```
chienandalu: :thinking: tests assertions should be deterministic

ernestotejeda:
@chienandalu could you review again?

event_mail/tests/test_event_mail.py
```diff
@@ -86,7 +87,7 @@ def test_event_template_no_config(self):
         event = self.env["event.event"].create(vals)
         self.assertEqual(
             len(event.event_mail_ids),
-            0,
+            3,
             "Event Mail: mails scheduler no created for this event",
```
chienandalu: But then, was the test wrong before? :thinking:

ernestotejeda: Here the default value of the `event_mail_ids` field in events is being testing when no template is defined in the res.config.setting. In v13, the event was created with an empty `event_mail_ids`, but in v15, it takes what the `_default_event_mail_type_ids` method of `event.type` returns  https://github.com/OCA/OCB/blob/15.0/addons/event/models/event_event.py#L27 and creates 3 records for `event_mail_ids`.

chienandalu:
please review @stefan-tecnativa

event_mail/tests/test_event_mail.py
```diff
@@ -0,0 +1,110 @@
...
+        vals = {
+            "name": "Event test",
+            "date_begin": "2017-05-01",
+            "date_end": "2017-06-01",
+            "auto_confirm": False,
+        }
+        event = self.env["event.event"].create(vals)
+        self.assertEqual(
+            len(event.event_mail_ids),
+            3,
```
stefan-tecnativa, line 90:
```suggestion
            len(event.event_type_id._default_event_mail_type_ids()),
```

As the event hasn't any type defined by default, we could compare the mails scheduled for the event with those that are scheduled by default when the type is not defined for the event. @chienandalu what do you think about this?

event_mail/readme/USAGE.rst
```diff
@@ -0,0 +1,9 @@
+To use this module, you need to:
+
+#. Go to *Events > Configuration > Mail Templates* and create all templates
+   that you need.
+#. Go to *Events > Settings* and select the default template in
+   "Template Mail Scheduler" field.
+#. Go to *Events > Events* and create one. All mails schedulers has been
+   created. Also you can select other template in "Mail Template Scheduler"
+   field on "Email Schedule" tab.
```
stefan-tecnativa:
```suggestion
   field on "Communication" tab.
```

stefan-tecnativa (commented):
Functionally tested, LGTM. Just check the comments.

ernestotejeda:
@stefan-tecnativa changes done

event_mail/tests/test_event_mail.py
chienandalu: I'm ok with the suggestion :+1:

chienandalu (approved):
Thanks everyone :)

ivantodorovich (approved):
LG

PR #314, opened 2023-06-08
[13.0] [ADD] website_event_sale_check_order_before_payment

Camille0907:
Add a way to check at the very last moment before the payment that the order is still valid, i.e. that the order has not been cancelled, and that there are still enough available seats / tickets for the selected event.


website_event_sale_check_order_before_payment/controllers/main.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+class WebsiteSale(website_sale.WebsiteSale):
+    @http.route(
+        ["/shop/payment/order_validity"], type="json", auth="public", website=True
+    )
+    def check_order_validity(self):
+        order = request.website.sale_get_order()
+        if order:
+            if order.state != "draft":
```
yankinmax:
```suggestion
        if order and order.state != "draft":
```

website_event_sale_check_order_before_payment/controllers/main.py
```diff
@@ -0,0 +1,85 @@
...
+                return {
+                    "order_is_valid": False,
+                    "invalid_order_message": request.env.company.cancelled_order_message,
+                }
+
+            order_lines_by_event = (
```
yankinmax, line 24: I wonder if we could avoid calling twice `read_group` and changing only the field to group on.

yankinmax (approved):

website_event_sale_check_order_before_payment/controllers/main.py
Camille0907: I'm not sure of it... A big part of the function is in the `if order` statement, and not in the `if order.state != "draft"` (line 24 to 64)

website_event_sale_check_order_before_payment/controllers/main.py
Camille0907: At first sight I don't see how... But I'll try if someone has solutions :)

website_event_sale_check_order_before_payment/static/src/js/event_payment_form.js
```diff
@@ -0,0 +1,33 @@
+odoo.define(
+    "website_event_sale_check_order_before_payment.event_payment_form",
+    function(require) {
+        "use strict";
+
+        var paymentForm = require("payment.payment_form");
```
ivantodorovich:
```suggestion
        const paymentForm = require("payment.payment_form");
```

nitpicking: `var` is old-js by convention

website_event_sale_check_order_before_payment/static/src/js/event_payment_form.js
```diff
@@ -0,0 +1,33 @@
...
+                    params: [],
+                });
+                if (order_validity_data.order_is_valid) {
+                    return _super.apply(this, arguments);
+                }
+                this.displayError(
+                    _t("The payment can't be processed"),
+                    _t(order_validity_data.invalid_order_message || "Unexpected error")
+                );
+                return;
```
ivantodorovich:
```suggestion
```
nitpicking

website_event_sale_check_order_before_payment/static/src/js/event_payment_form.js
```diff
@@ -0,0 +1,33 @@
...
+            /**
+             * @override
+             */
+            // eslint-disable-next-line no-unused-vars
+            payEvent: async function(ev) {
+                // This._super only works before any await operation, so store it here
+                const _super = this._super;
+                const order_validity_data = await this._rpc({
+                    route: "/shop/payment/order_validity",
+                    params: [],
```
ivantodorovich:
```suggestion
```
nitpicking: Isn't that the default? (not sure)

website_event_sale_check_order_before_payment/controllers/main.py
```diff
@@ -0,0 +1,85 @@
...
+        ["/shop/payment/order_validity"], type="json", auth="public", website=True
+    )
+    def check_order_validity(self):
+        order = request.website.sale_get_order()
+        if order:
+            if order.state != "draft":
+                return {
+                    "order_is_valid": False,
+                    "invalid_order_message": request.env.company.cancelled_order_message,
+                }
```
ivantodorovich:
```suggestion
                return {
                    "valid": False,
                    "message": request.env.company.cancelled_order_message,
                }
```

Nitpicking: I'd reduce the verbosity of the dict key names, like this.

website_event_sale_check_order_before_payment/controllers/main.py
ivantodorovich: Is it really cancelled if the `order.state != "draft"`?
Couldn't the order be `sent`, for example?

What's wrong with actually checking if it's cancelled? Are you worried about the other states?
```suggestion
            if order.state == "cancel":
```

website_event_sale_check_order_before_payment/controllers/main.py
```diff
@@ -0,0 +1,85 @@
...
+import odoo.addons.website_sale.controllers.main as website_sale
+
+_logger = logging.getLogger(__name__)
+
+
+class WebsiteSale(website_sale.WebsiteSale):
+    @http.route(
+        ["/shop/payment/order_validity"], type="json", auth="public", website=True
+    )
+    def check_order_validity(self):
```
ivantodorovich:
```suggestion
        ["/shop/check_before_payment"], type="json", auth="public", website=True
    )
    def check_before_payment(self):
```

Nitpicking: I'd use this route instead. "validity" somehow doesn't seem right, makes me think of an offer expiration


💡 Idea: since this module performs some other checks not really related to events, it could be split in two: `e-commerce/website_sale_check_before_payment` and `event/website_event_sale_check_before_payment`, the later depending on the other. Totally optional and/or it could even be a mention in the ROADMAP

website_event_sale_check_order_before_payment/models/res_company.py
```diff
@@ -0,0 +1,28 @@
...
+        "event page to register to another event.",
+        translate=True,
+    )
+    no_more_ticket_message = fields.Char(
+        help="Message displayed, if the last tickets have been sold before the "
+        "current payment process. (Availability per ticket)",
+        default="The last available tickets have just been sold. Please go back to "
+        "our event page to register to another event.",
+        translate=True,
+    )
```
ivantodorovich, line 28: Configurable error messages seem overkill IMO 🤔 
Imagine if all error messages in Odoo were configurable 🤯 

Honestly I'd remove this and stick to hardcoded messages.

But if you keep them, I guess the right model to store them isn't the `res.company`, but rather the `website` model. Example: https://github.com/OCA/e-commerce/blob/3b51c44cdd585868665e5e051f351827822c2bd0/website_sale_checkout_skip_payment/models/res_config_settings.py#L9-L13

website_event_sale_check_order_before_payment/controllers/main.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def _check_ticket_availability(self, ticket, ticket_qty):
+        if ticket.seats_availability == "limited":
+            if ticket.seats_available < ticket_qty:
+                return False
+        return True
+
+    def _cancel_cart_with_unavailable_seats(self, cart):
+        try:
+            with request.env.cr.savepoint():
```
ivantodorovich: What's the savepoint for ?

ivantodorovich (approved):
Purely optional comments/recommendations, feel free to ignore

ivantodorovich:
It would be really really nice to have unit tests, though

website_event_sale_check_order_before_payment/static/src/js/event_payment_form.js
Camille0907: Yes it is

website_event_sale_check_order_before_payment/controllers/main.py
Camille0907: You're right, checking cancel state should be enough

website_event_sale_check_order_before_payment/controllers/main.py
Camille0907: I agree, for now I'll mention it in the roadmap

website_event_sale_check_order_before_payment/controllers/main.py
Camille0907: Bad copy/paste of misunderstood code :wink:

website_event_sale_check_order_before_payment/static/src/js/website_sale_payment.js
```diff
@@ -0,0 +1,17 @@
...
+
+    const publicWidget = require("web.public.widget");
+
+    publicWidget.registry.WebsiteSalePayment.include({
+        /**
+         * @override
+         */
+        start: function() {
+            const res = this._super.apply(this, arguments);
+            this.$payButton = $("button#o_payment_form_check_order_and_pay");
```
ivantodorovich, line 15: What's this for? 🤔 
Maybe it's worth adding a docstring here

Camille0907: Yep, I'll add one :)
The standard button is enabled/disabled according whether the cgv checkbox is checked, and all the logic is done based on this `$payButton`. Overriding the content of the attribute allows to keep the standard behavior related to cgv_checkbox on the button with the new ID

ivantodorovich: Alright then ;) thanks!

ivantodorovich (approved):
TODO: squash commits to have only 1 adding the module

PR #316, opened 2023-06-14
[16.0][MIG] event_mail: Migration to version 16.0

stefan-tecnativa:
cc @Tecnativa TT43665

@ernestotejeda @chienandalu please review!


event_mail/views/event_view.xml
```diff
@@ -9,6 +9,7 @@
             <xpath expr="//field[@name='event_mail_ids']" position="before">
                 <field
                     name="event_mail_template_id"
+                    groups="base.group_user"
```
chienandalu, line 12: Why is this needed?

stefan-tecnativa: When trying to install the module, it gave me an error complaining that 
`Field 'seats_expected' used in attrs ({'readonly': [('seats_expected', '>', 0)]}) is restricted to the group(s) base.group_user.
`

chienandalu: Ok, thanks

chienandalu (approved):

PR #322, opened 2023-06-23
[16.0][MIG] partner_event: Migration to version 16.0

kwoychesko:
Biggest change is that _onchange_partner_id was removed in v16 core code so we no longer super it here. Instead, we call the _synchronize_partner_values method, which was the main job of the old _onchange_partner_id.


partner_event/models/event_registration.py
```diff
@@ -64,7 +64,13 @@ def _onchange_partner_id(self):
             get_attendee_partner_address = {
                 "get_attendee_partner_address": self.attendee_partner_id,
             }
-            return super(
-                EventRegistration, self.with_context(**get_attendee_partner_address)
-            )._onchange_partner_id()
-        return super(EventRegistration, self)._onchange_partner_id()
+            # onchange for partner_id removed in v16 core - including functionality here
```
chienandalu, line 67: Why not changing the whole logic to a compute readonly=False store=True as it's donde with the other fields? (e. g: https://github.com/OCA/OCB/blob/4860a93e28a82003f881e5c21ddc2dd4229e14a8/addons/event/models/event_registration.py#L62-L69

partner_event/__manifest__.py
```diff
@@ -9,13 +9,13 @@
...
     "name": "Link partner to events",
-    "version": "15.0.1.0.0",
+    "version": "16.0.1.0.0",
     "category": "Marketing",
     "author": "Tecnativa," "Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/event",
     "development_status": "Production/Stable",
     "license": "AGPL-3",
-    "depends": ["event"],
+    "depends": ["base", "event", "website"],
```
chienandalu, line 18: Why these new deps?

chienandalu (changes requested):
Thanks :) Some comments:

- Change your commit message following the guidelines

partner_event/__manifest__.py
kwoychesko: Thank you for reviewing! It's as a result of some errors raised when executing the tests. Messages like the following occur:

```
odoo.addons.base.models.ir_qweb.QWebException: Error while render the template
Exception: Unallowed to fetch files from addon website
Template: web.report_layout
```
In retrospect, adding the "website" dependency fixes the error and allows the tests to run, and probably "base" is unnecessary.

I'm new to migrating addons, so if you have a better way to handle this I'd appreciate any input!

partner_event/models/event_registration.py
kwoychesko: I'm sorry I don't understand the question. Are you able to expand any further? Do you mean: change the definition of the attendee_partner_id field to be more like the definition of the email field in the core code? E.g.:

`email = fields.Char(string='Email', compute='_compute_email', readonly=False, store=True, tracking=11)
`

pedrobaeza:
Please include #340

carolinafernandez-tecnativa:
@kwoychesko could you please include https://github.com/OCA/event/pull/340 ?
Thnks

pedrobaeza:
Superseded by #345

PR #325, opened 2023-06-30
[16.0][MIG] event_session: Migration to version 16.0

stefan-tecnativa:
cc @Tecnativa TT43670

@chienandalu @CarlosRoca13 please review!


event_session/models/event_registration.py
```diff
@@ -53,7 +53,9 @@ def _check_seats_limit(self):
                 and session.seats_available < (1 if rec.state == "draft" else 0)
             ):
                 raise ValidationError(_("No more seats available for this session."))
-        return super(EventRegistration, regular_records)._check_seats_limit()
+        return super(
+            EventRegistration, regular_records
+        ).event_id._check_seats_availability()
```
CarlosRoca13: Why are u calling `super()` if you are calling a method of other model?

If this method does not exist on the model probably it's not needed.

Please check it.

event_session/models/event_session.py
```diff
@@ -237,6 +237,7 @@ def _compute_seats_expected(self):
 
     @api.depends("date_tz", "date_begin")
     def _compute_date_begin_located(self):
+        # import wdb;wdb.set_trace()
```
CarlosRoca13:
```suggestion
```

CarlosRoca13 (changes requested):

CarlosRoca13 (approved):

event_session/models/event_event.py
```diff
@@ -102,12 +102,12 @@ def _compute_date_end(self):
             if data.get(rec.id):
                 rec.date_end = data.get(rec.id)
 
-    def _check_seats_limit(self):  # pragma: no cover
+    def _check_seats_availability(self):  # pragma: no cover
```
chienandalu: you're changing the method signature: https://github.com/odoo/odoo/blame/16.0/addons/event/models/event_event.py#L555

event_session/views/event_session.xml
```diff
@@ -344,29 +344,26 @@
...
+                                        t-if="record.date_begin.raw_value !== record.date_end.raw_value"
                                     >
-                                        <i
+                                            <i
                                             class="fa fa-arrow-right o_event_fontsize_09"
                                             title="End date"
                                         />
-                                        <t
-                                            t-esc="moment(record.date_end.raw_value).format('D MMM')"
+                                            <t
```
chienandalu: These indentation changes look unnecessary

event_session/views/event_session.xml
```diff
@@ -0,0 +1,520 @@
...
+                <templates>
+                    <t t-name="kanban-box">
+                        <div
+                            t-attf-class="d-flex flex-column p-0 oe_kanban_card oe_kanban_global_click"
+                        >
+                            <div
+                                class="o_kanban_content p-0 m-0 position-relative row d-flex flex-fill"
+                            >
+                                <div
+                                    class="col-3 bg-primary p-2 text-center d-flex flex-column justify-content-center"
```
chienandalu:
```suggestion
                                    class="col-3 text-bg-primary p-2 text-center d-flex flex-column justify-content-center"
```

event_session/models/event_event.py
```diff
@@ -102,12 +102,12 @@ def _compute_date_end(self):
...
         # OVERRIDE to ignore this constraint for event with sessions
         # Seat availability is checked on each session, not here.
         session_records = self.filtered("use_sessions")
         regular_records = self - session_records
-        return super(EventEvent, regular_records)._check_seats_limit()
+        return super(EventEvent, regular_records)._check_seats_availability()
```
chienandalu: But you should check the session seats availability somehow, and this makes me think that the tests aren't complete. Please check how this has changed in v16 in core and it for the sessions.

chienandalu (changes requested):

event_session/models/event_event.py
ivantodorovich: Note: For events with sessions, the limit is applied on the `event.session` level. The `event.event` is something like a "template"

event_session/models/event_session.py
```diff
@@ -206,7 +206,7 @@ def _compute_seats(self):
                 AND state IN %s
                 GROUP BY session_id, state
             """
-            self.env["event.registration"].flush(["session_id", "state"])
+            self.env["event.registration"]._flush(["session_id", "state"])
```
ivantodorovich:
```
self.env["event.registration"].flush_model(["session_id", "state"])
```

event_session/tests/test_event_session.py
```diff
@@ -202,7 +191,7 @@ def test_session_seats(self):
             ValidationError, "No more available seats."
         ), self.cr.savepoint():
             registration.action_confirm()
-            registration.flush()
+            registration._flush()
```
ivantodorovich: `registration.flush_recordset()`

event_session/models/event_event.py
chienandalu: Thanks @ivantodorovich :slightly_smiling_face:

event_session/tests/test_event_session.py
```diff
@@ -356,10 +345,10 @@ def test_event_session_is_finished(self):
 
     def test_event_session_registrations_open(self):
         with freeze_time("2017-05-26 20:30:00"):
-            self.session.invalidate_cache(["event_registrations_open"])
+            self.session._invalidate_cache(["event_registrations_open"])
```
ivantodorovich: `self.session.invalidate_recordset`

event_session/tests/test_event_session.py
```diff
@@ -356,10 +345,10 @@ def test_event_session_is_finished(self):
...
             self.assertTrue(self.session.event_registrations_open)
         with freeze_time("2017-05-30 20:00:00"):
-            self.session.invalidate_cache(["event_registrations_open"])
+            self.session._invalidate_cache(["event_registrations_open"])
```
ivantodorovich: `self.session.invalidate_recordset`

event_session/models/event_session.py
```diff
@@ -0,0 +1,524 @@
...
+                ]
+                # Force recomputation of scheduled date
+                rec.event_mail_ids._compute_scheduled_date()
+
+    @api.model
+    def _read_group_stage_ids(self, stages, domain, order):  # pragma: no cover
+        return self.env["event.event"]._read_group_stage_ids(stages, domain, order)
+
+    @api.constrains("seats_max", "seats_available", "seats_limited")
+    def _check_seats_limit(self):
```
ivantodorovich: Could you update this method so that it's named the same as, and behaves the same as, [the `_check_seats_availability` on `event.event`](https://github.com/odoo/odoo/blob/d44106309c4a11fc3c0dca4ed8457e1f27bd1527/addons/event/models/event_event.py#L555) ?

The general idea is the `event.session` should act almost exactly as the core `event.event` for regular events. So most of the methods, fields and constrains here are adaptations of what you find in `event.event`

ivantodorovich (commented):
Hey, thanks for the work done!

I did a first quick review, but I'll do a second pass asap

event_session/models/event_session.py
ivantodorovich: > The general idea is the event.session should act almost exactly as the core event.event for regular events. So most of the methods, fields and constrains here are adaptations of what you find in event.event

It'd be great if you can check the other fields / methods with this in mind, too 🙏🏻

chienandalu:
@stefan-tecnativa please include https://github.com/OCA/event/pull/328 once is merged

chienandalu:
> @stefan-tecnativa please include #328 once is merged

Already merged

ivantodorovich:
Without the "SAVE" button on form views, this mechanism to write on multiple sessions at once is a bit weird.
I feel it'll be confusing for the user to know what he's modifying because he doesn't know when the record is gonna be saved.

image

This mechanism was inspired by a simlar Odoo core feature in recurring Calendar Events. I checked to see how Odoo dealt with the dissapearence of the "SAVE" button there, and it seems nothing changed

image

Althuogh it's more commonly modified from the calendar view, where there's the "SAVE" button, so I guess they just don't care / didn't notice it

image


----

I'm not sure what the best thing to do is here, options are:

* **A. Keep it as-is**
* **B. Deprecate / remove this feature**
* **C. Find a better way, in terms of UX**

Opinions?

ping @gregory-moka

ivantodorovich (approved):
Apart from the topic mentioned above, I did a code and functional review and it all seems good ! Nice work!

event_session/models/event_event.py
```diff
@@ -102,12 +102,12 @@ def _compute_date_end(self):
...
                 rec.date_end = data.get(rec.id)
 
-    def _check_seats_limit(self):  # pragma: no cover
+    def _check_seats_availability(self, minimal_availability=0):  # pragma: no cover
         # OVERRIDE to ignore this constraint for event with sessions
         # Seat availability is checked on each session, not here.
         session_records = self.filtered("use_sessions")
         regular_records = self - session_records
-        return super(EventEvent, regular_records)._check_seats_limit()
+        return super(EventEvent, regular_records)._check_seats_availability()
```
chienandalu: You're inhibiting this argument
```suggestion
        return super(EventEvent, regular_records)._check_seats_availability(minimal_availability=minimal_availability)
```

event_session/models/event_registration.py
```diff
@@ -40,21 +54,6 @@ def _compute_event_end_date(self):
             else:
                 rec.event_end_date = rec.event_id.date_end
 
-    @api.constrains("session_id")
```
chienandalu: Keep the method where it was to minimize the listed changes

event_session/models/event_registration.py
```diff
@@ -24,6 +24,20 @@ class EventRegistration(models.Model):
     )
     event_end_date = fields.Datetime(related=None, compute="_compute_event_end_date")
 
+    @api.constrains("session_id")
+    def _check_seats_limit(self):
```
chienandalu: Anyway, is this needed at all having the same constraint at session level?

chienandalu (changes requested):

chienandalu:
> Without the "SAVE" button on form views, this mechanism to write on multiple sessions at once is a bit weird. I feel it'll be confusing for the user to know what he's modifying because he doesn't know when the record is gonna be saved.
> I'm not sure what the best thing to do is here, options are:
> 
> A. Keep it as-is
> B. Deprecate / remove this feature
> C. Find a better way, in terms of UX

I'd keep it, and a better way to handle UX could be to have those options folded by default (using bootstrap magic) so the user that knows what's doing can take advantage of them.

stefan-tecnativa:
First of all thanks you @ivantodorovich and @chienandalu for your time and suggestions. About what you were discussing, About what you were discussing, I feel that right now I can't give an opinion, so I prefer to let the experts comment 😉

chienandalu:
> About what you were discussing, I feel that right now I can't give an opinion, so I prefer to let the experts comment wink

Ok, summarize the discussion in the roadmap

stefan-tecnativa:
> > About what you were discussing, I feel that right now I can't give an opinion, so I prefer to let the experts comment wink
> 
> Ok, summarize the discussion in the roadmap

Understood

PR #332, opened 2023-07-11
[16.0][MIG] event_session_registration_multi_qty: Migration to version 16.0

stefan-tecnativa:
cc @Tecnativa TT43671

@pedrobaeza @CarlosRoca13 please review!

This PR depends on: #325


event_session_registration_multi_qty/models/event_session.py
```diff
@@ -16,8 +16,7 @@ def _compute_seats(self):
                 "seats_reserved": 0,
                 "seats_used": 0,
                 "seats_available": 0,
-                "seats_available_expected": 0,
-                "seats_available_pc": 0,
+                "seats_expected": 0,
```
pedrobaeza: You are not assigning after this any value to `seats_expected`, so that seems a mistake. And please test this new field.

stefan-tecnativa: I'm not getting it. Isn't the field being used to update the sessions fields in the last `session.update(vals)` ?? And the field is checked in the tests. Are you reffering to another field different than `seats_expected` that i haven't checked?

pedrobaeza: My bad, sorry.

pedrobaeza: Anyway, I see you don't need this initialization, as the value is always unconditionally set.

pedrobaeza (approved):

event_session_registration_multi_qty/models/event_session.py
stefan-tecnativa: Changed, thanks for the revision.

CarlosRoca13 (approved):
LGTM :+1:

PR #333, opened 2023-07-13
[16.0][MIG] event_sale_session: Migration to version 16.0

stefan-tecnativa:
cc @Tecnativa TT43669

@CarlosRoca13 @pedrobaeza please review!


pedrobaeza:
Remember to squash administrative commits.

event_sale_session/static/src/js/sale_product_field.esm.js
CarlosRoca13, line 1: Remove this file

CarlosRoca13 (changes requested):

event_sale_session/static/src/js/sale_product_field.esm.js
```diff
@@ -0,0 +1,53 @@
...
+patch(
+    SaleOrderLineProductField.prototype,
+    "event_sale_session.SaleOrderLineProductField",
+    {
+        /**
+         * This method is completely overwritten because odoo is limiting the fields that
+         * can be written to event_id and event_ticket_id. This can be deleted by changing
+         * the following:
+         * https://github.com/odoo/odoo/blob/fa142764b2f583793047f794bbc102e6e0807c31/addons/event_sale/static/src/js/sale_product_field.js#L52-L55
+         * TODO: Remove if this PR is merged: https://github.com/odoo/odoo/pull/128437
```
ivantodorovich: https://github.com/odoo/odoo/pull/128437 is getting merged 🥳

event_sale_session/__manifest__.py
```diff
@@ -22,7 +22,7 @@
     "demo": ["demo/event_session.xml"],
     "assets": {
         "web.assets_backend": [
-            "event_sale_session/static/src/js/EventConfiguratorFormController.js",
+            "event_sale_session/static/src/js/*.esm.js",
```
ivantodorovich, line 26: Any reason why this commit is not merged with the `Migration to 16.0` commit?
It sounds to me this change is required for the migration, so it makes sense to have it there 🤔

event_sale_session/models/sale_order_line.py
```diff
@@ -25,16 +26,16 @@ def _onchange_event_id(self):
             self.event_session_id = event.session_ids[0]
         elif not event.use_sessions or event != self.event_session_id.event_id:
             self.event_session_id = False
-        return super()._onchange_event_id()
+        return super()._compute_event_ticket_id()
```
ivantodorovich: ⚠️ : This doesn't seem right.

This method is named `_onchange_event_id` but it's calling super `_compute_event_ticket_id`

event_sale_session/reports/sale_report.py
```diff
@@ -27,10 +27,25 @@ class SaleReport(models.Model):
...
     )
 
-    def _query(self, with_clause="", fields=None, groupby="", from_clause=""):
-        fields = fields or {}
-        fields["event_id"] = ", l.event_id as event_id"
-        fields["event_ticket_id"] = ", l.event_ticket_id as event_ticket_id"
-        fields["event_session_id"] = ", l.event_session_id as event_session_id"
-        groupby += ", l.event_id, l.event_ticket_id, l.event_session_id"
-        return super()._query(with_clause, fields, groupby, from_clause)
+    def _query(self):
```
ivantodorovich: Rather than a full overwrite on `_query`, I think we want to override `_select_sale` and `_group_by_sale` instead.

event_sale_session/models/sale_order_line.py
ivantodorovich: After checking, I think we'd want to align with what Odoo's made for `event_ticket_id`, [making it a computed field (store=True, readonly=False, precompute=True)](https://github.com/odoo/odoo/blob/68c6416c03c31a04b4458f8a64886215d14675ba/addons/event_sale/models/sale_order.py#L69-L72).

So, in our case, `event_session_id` should be a computed method with its own `_compute_event_session_id`

event_sale_session/models/sale_order_line.py
```diff
@@ -0,0 +1,43 @@
...
+            self.event_session_id = event.session_ids[0]
+        elif not event.use_sessions or event != self.event_session_id.event_id:
+            self.event_session_id = False
+        return super()._compute_event_ticket_id()
+
+    @api.onchange("event_session_id")
+    def _onchange_event_session_id(self):
+        # We call this to force update the default name
+        # Simliar to :meth:`~_onchange_event_ticket_id` in core.
+        self._compute_name()
```
ivantodorovich: To align with [what's done now in core for `event_ticket_id` field](https://github.com/odoo/odoo/blob/68c6416c03c31a04b4458f8a64886215d14675ba/addons/event_sale/models/sale_order.py#L140-L146)

```suggestion
    @api.depends("event_session_id")
    def _compute_name(self):
        # OVERRIDE to add the compute method dependency.
        #
        # Simliar to what's done in core for the `event_ticket_id` field.
        # See :meth:`~_get_sale_order_line_multiline_description_sale`
        super()._compute_name()
```

event_sale_session/models/sale_order_line.py
stefan-tecnativa: Thanks for all the comments 😃.
I've checked it and made `event_session_id` a compute field, similar to `event_ticket_id`. Have I done it right?

event_sale_session/models/sale_order_line.py
```diff
@@ -13,28 +13,42 @@ class SaleOrderLine(models.Model):
...
         string="Session",
         comodel_name="event.session",
+        compute="_compute_event_session_id",
         domain="[('event_id', '=', event_id)]",
+        store=True,
+        readonly=False,
+        precompute=True,
     )
 
+    @api.depends("event_session_id")
```
CarlosRoca13: It should depens on event_id

event_sale_session/models/sale_order_line.py
CarlosRoca13: If yo make the computed field you don't need to overwrite the _compute_event_ticket_id

stefan-tecnativa: Got it, can you check it please?

event_sale_session/models/sale_order_line.py
```diff
@@ -13,28 +13,33 @@ class SaleOrderLine(models.Model):
...
-            self.event_session_id = event.session_ids[0]
-        elif not event.use_sessions or event != self.event_session_id.event_id:
-            self.event_session_id = False
-        return super()._onchange_event_id()
+    @api.depends("event_id")
+    def _compute_event_session_id(self):
+        event_lines = self.filtered("event_id")
+        (self - event_lines).event_session_id = False
+        for line in event_lines:
+            if line.event_id != line.event_session_id.event_id:
```
CarlosRoca13:
```suggestion
            if not line.event_id.use_sessions or line.event_id != line.event_session_id.event_id:
```
I think that you need to preserve the fist part of the condition

event_sale_session/static/src/js/EventConfiguratorFormController.esm.js
```diff
@@ -0,0 +1,36 @@
...
+    Copyright 2023 Tecnativa - Carlos Roca
+    @author Iván Todorovich <ivan.todorovich@gmail.com>
+    License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+*/
+import {RelationalModel} from "@web/views/relational_model";
+import {patch} from "@web/core/utils/patch";
+
+patch(RelationalModel.prototype, "event_sale_session.RelationalModel", {
+    setup(params, {action}) {
+        // Trick to keep model as the main EventConfiguratorRelationalModel is not exported
```
ivantodorovich, line 13: Most probably Odoo would accept a PR making both `EventConfiguratorRelationalModel` and `EventConfiguratorRecord` exportables.

event_sale_session/models/sale_order_line.py
stefan-tecnativa: Good catch, thanks.

ivantodorovich (commented):
In `16.0` odoo added this ["event sale report"](https://github.com/odoo/odoo/blob/16.0/addons/event_sale/report/event_sale_report.py).

I think it'd be cool to add the sessions in there. If you can't do it right now coudl you mention this in the roadmap?

stefan-tecnativa:
> In `16.0` odoo added this ["event sale report"](https://github.com/odoo/odoo/blob/16.0/addons/event_sale/report/event_sale_report.py).
> 
> I think it'd be cool to add the sessions in there. If you can't do it right now coudl you mention this in the roadmap?

Sure, I will take a look at it.

stefan-tecnativa:
@ivantodorovich I have added a commit with the extension of the `event.sale.report`. Can you check it please? If its okay, should I leave the commit or squash it with the migration one?

ivantodorovich:
Hello @stefan-tecnativa ! The report looks good!
I've just tested it on runboat 👍🏻 

---

What's not working properly for me is the event selector widget. Maybe our runboat's Odoo is not updated yet (and so it doesn't include the PR you made) ? Can you confirm it works properly on your side?

ivantodorovich:
> should I leave the commit or squash it with the migration one?

Either way is find IMO. It's an improvement not required for the migration, so it can stay as a separate atomic commit.

CarlosRoca13:
> Hello @stefan-tecnativa ! The report looks good! I've just tested it on runboat 👍🏻
> 
> What's not working properly for me is the event selector widget. Maybe our runboat's Odoo is not updated yet (and so it doesn't include the PR you made) ? Can you confirm it works properly on your side?
> 
> 

Hi! @ivantodorovich 
I think it is because the change is not yet included in OCB. 
https://github.com/OCA/OCB/blob/16.0/addons/event_sale/static/src/js/sale_product_field.js#L53-L54

stefan-tecnativa:
> 

Hi @ivantodorovich . I can confirm that in my local it is working properly. 

event_sale_session

ivantodorovich:
That makes sense, thanks for checking!

ivantodorovich (approved):
This is looking great to me 👍🏻  Thanks for the good work!

CarlosRoca13 (approved):

PR #335, opened 2023-08-02
[16.0][IMP] event_sale_session: Use EventConfiguratorRecord instead of RelationalModel

CarlosRoca13:
cc @Tecnativa TT43669

@ivantodorovich  @chienandalu I've been looking at options to extend the doAction method, but I haven't found anything that would work for me... If you can think of a better way to extend the doAction of the model's action object, it will be welcome.


chienandalu (approved):
:+1:

event_sale_session/static/src/js/EventConfiguratorFormController.esm.js
```diff
@@ -5,32 +5,33 @@
...
+                }
+                return model.oldDoAction(actionRequest, options);
             }
-            return model.oldDoAction(actionRequest, options);
-        }
-        this.action.doAction = EventConfiguratorDoAction;
-    },
-});
+            this.action.doAction = EventConfiguratorDoAction;
+        },
```
ivantodorovich: This is what I normally do on cases like this:

```js
patch(
    EventConfiguratorRelationalModel.prototype,
    "event_sale_session.EventConfiguratorRelationalModel",
    {
    /**
     * @override
     */
    async save() {
        const doAction = this.model.doAction;
        this.model.doAction = (actionRequest, options = {}) => {
            actionRequest.infos.eventConfiguration.event_session_id = this.data.event_session_id;
            return doAction(actionRequest, options);
        }
        const res = this._super.apply(this, arguments);
        this.model.doAction = doAction;
        return res;
    }
}
```

Basically, temporarily monkey-patching the method

ivantodorovich (approved):

event_sale_session/static/src/js/EventConfiguratorFormController.esm.js
CarlosRoca13: Let me try it :smile: thanks

CarlosRoca13:
I have made the changes to follow @ivantodorovich 's style.

Thank you very much, this way it has been much simpler.

chienandalu (approved):

PR #340, opened 2023-08-30
[15.0][IMP] partner_event: Do registration creation in batch + remove savepoint

pedrobaeza:
For speeding up the process. Specially the removal of the savepoints makes that Postgres don't go slow as hell. Check https://www.cybertec-postgresql.com/en/subtransactions-and-performance-in-postgresql/ for some rationale. The problem happens even releasing the previous savepoints.

As this module is very old, the reason for doing it this way and skip registrations with error is not fully known, but the view mentions about registration duplication, so let's avoid it and handle later other possible exceptions.

@Tecnativa TT26694

### DIFF
--- a/partner_event/models/event_registration.py
+++ b/partner_event/models/event_registration.py
@@ -1,8 +1,8 @@
-# Copyright 2014 Tecnativa S.L. - Pedro M. Baeza
-# Copyright 2015 Tecnativa S.L. - Javier Iniesta
-# Copyright 2016 Tecnativa S.L. - Antonio Espinosa
-# Copyright 2016 Tecnativa S.L. - Vicent Cubells
-# Copyright 2020 Tecnativa S.L. - Víctor Martínez
+# Copyright 2015 Tecnativa - Javier Iniesta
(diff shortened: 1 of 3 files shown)


ivantodorovich (approved):
👍🏻 

You can probably remove the `errors` field, too.. it seems it was only used to store the error messages you've just removed.

victoralmau (approved):

chienandalu (approved):

pedrobaeza:
While removing the `errors` field stuff, I see that in the view there's a mention saying about duplicates. Although there's no constraint for avoiding such duplication, there's not too much sense to register several times the same contact as attendee, so I'm now skipping these duplicates.

Do you see it correct? Should I put a constraint as well?

ivantodorovich:
👍🏻  I think it's correct like this. If the partner is already registered, it make sense to simply skip it
The intention is to ensure the partners are registered, I guess

pedrobaeza:
OK, let's go for now improving this part:

/ocabot merge patch

PR #342, opened 2023-09-19
[15.0][FIX] partner_event: compatibility with event_sale

chienandalu:
When we confirm a sale order, the registrations are precreated without the email info, so we were missing the chance to get the info for the partner attendant.

Now we compute it on the write as well.

cc @Tecnativa TT45149

### DIFF
--- a/partner_event/models/event_registration.py
+++ b/partner_event/models/event_registration.py
@@ -27,26 +27,34 @@ class EventRegistration(models.Model):
             "phone": vals.get("phone", False),
         }
 
+    def _update_attendee_partner_id(self, vals):
+        if not vals.get("attendee_partner_id") and vals.get("email"):
+            Partner = self.env["res.partner"]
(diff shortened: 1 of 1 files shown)


partner_event/models/event_registration.py
```diff
@@ -27,26 +27,34 @@ def _prepare_partner(self, vals):
...
-                    # Create partner
-                    attendee_partner = Partner.sudo().create(
-                        self._prepare_partner(values)
-                    )
-                values["attendee_partner_id"] = attendee_partner.id
-        return super(EventRegistration, self).create(vals_list)
+            self._update_attendee_partner_id(values)
+        return super().create(vals_list)
+
+    def write(self, values):
```
pedrobaeza: Use the standard signature:

```suggestion
    def write(self, vals):
```

victoralmau (approved):

PR #351, opened 2023-11-08
[15.0][FIX] partner_event: NewId compatibility

chienandalu:
When the compute is triggered from a NewId the domain will fail when it's evaluated as expression parser can't deal with it.

cc @Tecnativa TT45858


pedrobaeza (approved):
/ocabot merge patch

PR #352, opened 2023-11-13
[15.0][FIX] event_registration_partner_unique: avoid constraint on merge

chienandalu:
When a user merges two contacts an error will raise if there are event with a unique partner constraint when both partner were atendees. With these changes, we let the user skip the constraint.

cc @Tecnativa TT45921


pedrobaeza (changes requested):
Letting the duplicate registration is not correct IMO. What should be done is that only one registration is left (the oldest one).

chienandalu:
I don't think is that easy. What happens if both registrations were valid? The unique flag could be set after two resgistrations had the same attendee and those two would be still valid (e.g.: booked by the same attendee to different contacts)

pedrobaeza:
I know it's not easy, but the duplication is simply not correct. You should forbid it or merge it. Let's do this: add an option and merge contact dialog (disabled by default) to merge them. If not enabled, forbid the merging. And let's put the default True in our customer.

chienandalu:
Check it now

pedrobaeza:
This is not what I was talking about. Registrations should be merged as well (or the newer one being removed).

chienandalu:
Changed to get rid of the newest duplicated attendees

event_registration_partner_unique/models/event.py
```diff
@@ -42,6 +57,8 @@ def _check_forbid_duplicates(self):
                         ),
                     }
                 )
+        # Call with skipping context to avoid triggering the constraint again
+        dupes_to_unlink.with_context(skip_registration_partner_unique=True).unlink()
```
pedrobaeza: Why doing this here instead on the wizard? This way, it's less readable and non self-contained.

chienandalu: Do you mean the unlinking or the whole dupes evaluation?

pedrobaeza: Everything. The constraint shouldn't do anything except constraining.

chienandalu: Ok, I retailored the code a little bit to be able to do the attendee cleanup in the wizard

chienandalu:
Can you review again @pedrobaeza ?

event_registration_partner_unique/wizards/base_partner_merge.py
```diff
@@ -0,0 +1,33 @@
...
+            return super().action_merge()
+        # Skip the constraints and do the partner merge first
+        res = super(
+            BasePartnerMergeAutomaticWizard,
+            self.with_context(skip_registration_partner_unique=True),
+        ).action_merge()
+        # Now let's merge the attendees
+        dupes_to_unlink = self.env["event.registration"]
+        partner_registrations = self.dst_partner_id.event_registration_ids
+        for attendee, dupes in partner_registrations._find_duplicated_attendees():
```
pedrobaeza, line 26: **question:** Why not doing this before calling `super`?

event_registration_partner_unique/wizards/base_partner_merge.py
```diff
@@ -0,0 +1,33 @@
...
+            # Keep the oldest one -> min id will be
+            event_attendees_for_partner = attendee + dupes
+            attendee_to_keep = min(event_attendees_for_partner, key=lambda x: x.id)
+            dupes_to_unlink += event_attendees_for_partner - attendee_to_keep
+        # Call with skipping context to avoid triggering the constraint again
+        dupes_to_unlink.with_context(skip_registration_partner_unique=True).unlink()
```
pedrobaeza, line 32: **question:** Why this is requiring the skip context? There shouldn't be duplicates once unlinked, or is it trigger before unlinking? If so, please be more explicit in the comment.

event_registration_partner_unique/wizards/base_partner_merge.py
chienandalu: Before calling `super` they won't be duplicated as they'll belong to diferent partners

event_registration_partner_unique/wizards/base_partner_merge.py
chienandalu: > There shouldn't be duplicates once unlinked

There's a constraint trigger for `registration_ids` at `event.event` level, so I want to make sure we don't perform the check twice

event_registration_partner_unique/wizards/base_partner_merge.py
pedrobaeza: Well, on my approach, you search for the duplicated saying if the partner is one of thos to be merged, but OK for not disturbing more.

pedrobaeza (approved):

victoralmau (approved):

chienandalu:
@pedrobaeza merge?

chienandalu:
ping @pedrobaeza

PR #359, opened 2024-01-29
[15.0][OU] event_session: Migration script

chienandalu:
Preset the new use_sessions flag according to existing data

cc @Tecnativa TT46031

### DIFF
--- a/event_session/migrations/15.0.1.0.0/pre-migration.py
+++ b/event_session/migrations/15.0.1.0.0/pre-migration.py
@@ -1,0 +1,36 @@
+# Copyright 2012 Tecnativa - David Vidal
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from openupgradelib import openupgrade
+
+
+def _set_use_sessions_flag(env):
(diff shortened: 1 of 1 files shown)


ivantodorovich (approved):

event_session/migrations/15.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,36 @@
...
+            SELECT DISTINCT event_id FROM event_session
+        )
+    """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE event_type SET use_sessions = true
+        WHERE id in (
+            SELECT DISTINCT event_type_id FROM event_event WHERE use_sessions = true
```
pedrobaeza: Isn't possible to have a mix of events with/without sessions on the same type?

```suggestion
            SELECT DISTINCT event_type_id FROM event_event WHERE use_sessions
```

chienandalu: Sure, this is just a default for the events using this template.

pedrobaeza (approved):

PR #360, opened 2024-01-29
[16.0][IMP] event_session: improve UX in session tree

chienandalu:
- Hide tz: if our events are always in the same tz, it can be quite useless.
- Optionally display: event_id (known if you access from the button in the event)
- Add (hidden) the seats limited flag

cc @Tecnativa TT46031


ivantodorovich (approved):

victoralmau (approved):

pedrobaeza (approved):
/ocabot merge minor

PR #373, opened 2024-05-27
[16.0][IMP] event_sale_session: Recovery unconfirmed quantity event

carolinafernandez-tecnativa:
- Recovery functionality unconfirmed qty on event.session from v14

@Tecnativa
TT49353

@chienandalu @pedrobaeza


event_sale_session/models/event_event.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class EventEvent(models.Model):
+    _inherit = "event.event"
+
+    order_line_ids = fields.One2many(
+        comodel_name="sale.order.line",
+        inverse_name="event_id",
+        string="Sales Order Lines",
+    )
```
chienandalu: There's already a `sale_order_lines_ids` field provided by `event_sale`

event_sale_session/models/event_event.py
```diff
@@ -0,0 +1,44 @@
...
+        action = self.env["ir.actions.act_window"]._for_xml_id("sale.action_quotations")
+        sales = (
+            self.env["sale.order.line"]
+            .search([("event_id", "in", self.ids)])
+            .mapped("order_id")
+            .ids
+        )
+        action["domain"] = [("id", "in", sales), ("state", "in", ("draft", "sent"))]
+        action["context"] = {}
+        return action
```
chienandalu: This stuff doesn't belong to this module (I know it was, but it wasn't right).

carolinafernandez-tecnativa: Our customer needs the information to check from event form, that's why is necessary.

chienandalu: I know: but that doesn't mean that this code belongs here :smile It should either go to its own module or to custom code

event_sale_session/tests/test_event_sale_session.py
```diff
@@ -54,22 +57,6 @@ def test_sale_session(self):
...
-        line.event_id = self.event
-        self.assertFalse(line.event_session_id)
-        # Case 2: The event is a session event with only 1 session
-        (self.event.session_ids - self.session).active = False
-        line.event_id = self.event
-        self.assertEqual(line.event_session_id, self.session)
-        # Case 3: The event is not a session event, session should be unset
-        line.event_id = self.env.ref("event.event_0")
-        self.assertFalse(line.event_session_id)
-
```
chienandalu, line 72: Why removing this whole test?

carolinafernandez-tecnativa: Onchange is only applied when you are in form view

event_sale_session/models/event_event.py
ivantodorovich: I agree this "unconfirmed quantity" feature doesn't belong here. This module's purpose should be to implement the event sessions with at most the same features you find in standard Events module, not more.

Any other extra feature should be moved to a separate module. (e.g: `event_sale_session_unconfirmed_qty`? notice you could even have a `event_sale_unconfimed_qty` module that implements this in standard events, because the feature doesn't exist in standard)

event_sale_session/tests/test_event_sale_session.py
ivantodorovich: Sorry, I still don't understand the reason to remove it 😓 are you facing an issue with this test while implementing the new feature?

ivantodorovich (changes requested):
This should be in a separate `event_sale_session_unconfirmed_qty` module IMO

carolinafernandez-tecnativa:
@pedrobaeza what do you think to move this functionality to new module?
Also in v14 unconfirmed qty also was added to event.event and there was a button to check all sale orders with state in draft or sent associated to each event.
Maybe the new module should be called event_sale_unconfirmed_qty so we can add both functionalities, what do you think @ivantodorovich ?

ivantodorovich:
> Maybe the new module should be called event_sale_unconfirmed_qty so we can add both functionalities, what do you think @ivantodorovich ?

If it depends on `event_session` then it should be prefixed with it.

The idea to split them in `event_sale_unconfirmed_qty` and `event_sale_session_unconfirmed_qty` is to offer the same feature to people using the event modules but not the event_session module. But if you don't need the feature on the standard events then there's no point in developing it now, still.. it's nice to keep this in mind to structure the modules.

Take for example these two modules:
- event_registration_multi_qty
- event_session_registration_multi_qty

chienandalu:
@pedrobaeza please add "no stale" label

event_sale_session/tests/test_event_sale_session.py
pedrobaeza: Adding this uncovers a problem in a computed method, that has been fixed in an additional commit 9799cfb, so the test is restored.

pilarvargas-tecnativa (approved):

pedrobaeza (approved):
@ivantodorovich as the requested change has been fixed, I'm merging this for continuing with the fw-port, as we are even going now to 18.

Thank you all for your comments.

/ocabot merge minor

PR #374, opened 2024-05-28
[16.0][FIX] event_session: action permissions

chienandalu:
Quick fix to avoid access error over users with regular permissions

cc @Tecnativa TT49384

please check @pedrobaeza @carolinafernandez-tecnativa


pedrobaeza (approved):
Clear change:

/ocabot merge patch

PR #375, opened 2024-06-04
[16.0][ADD] website_event_membership_restriction: New module

pilarvargas-tecnativa:
This module allows you to restrict event registration to members only.

cc @tecnativa TT49482

@pedrobaeza @carolinafernandez-tecnativa please review

event

The same for logged in users whose associated commercial partner does not have membership_state ‘paid’ or ‘free’.


carolinafernandez-tecnativa:
Functional review LGTM, Translations are missing.

pilarvargas-tecnativa:
> Functional review LGTM, Translations are missing.

Done! thanks

carolinafernandez-tecnativa (approved):

website_event_membership_restriction/views/event_templates.xml
```diff
@@ -0,0 +1,133 @@
...
+        >
+            <attribute name="class">d-none</attribute>
+        </xpath>
+        <xpath
+            expr="//form//div[@id='o_wevent_tickets_collapse']//button"
+            position="after"
+        >
+            <button
+                type="submit"
+                t-att-class="'btn btn-primary o_wait_lazy_js a-submit ' + (event.allow_members_only and (not request.env.user or request.env.user.partner_id.commercial_partner_id.membership_state not in ['paid', 'free']) and 'disabled' or '')"
```
chienandalu: - Maybe you should check if the user isn't public better `not request.env.user._is_public()`
- Why checking the commercial_partner_id state? What if membership is given by an associaate?

pedrobaeza: @chienandalu do you mean a prorated membership?

website_event_membership_restriction/views/event_templates.xml
```diff
@@ -0,0 +1,103 @@
...
+            </button>
+            <t
+                t-if="event.allow_members_only and (not request.env.user or request.env.user.partner_id.commercial_partner_id.membership_state not in ['paid', 'free'])"
+            >
+                <a class="text-center small" href="/membership-info">Members only</a>
+            </t>
+        </xpath>
+    </template>
+
+    <template id="website_event_membership_info" name="Membership Info">
```
pedrobaeza: This should be a website.page with its `arch`, not a template to be rendered with a controller.

website_event_membership_restriction/readme/USAGE.rst
```diff
@@ -0,0 +1,5 @@
+
```
pedrobaeza: Remove this extra initial line.

website_event_membership_restriction/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,4 @@
+#. Go to *Events > Events*
+#. Create or select one event.
+#. Check or uncheck "Allow registration only to members".
+
```
pedrobaeza: Remove this empty line

website_event_membership_restriction/views/event_templates.xml
chienandalu: No, I mean that a contact could be assoiciated to other one membership: https://github.com/odoo/odoo/blob/5fa953a0886c0b45169fc894c6202ede2f68f700/addons/membership/models/partner.py#L13-L15

So a commercial partner could have some child contacts which inherit his membership and others that don't. Also contacts inheriting membership from other organization...

pedrobaeza: That's not the use case, but let's fallback to check if the commercial partner or the partner itself.

website_event_membership_restriction/models/event_event.py
```diff
@@ -0,0 +1,8 @@
+# Copyright 2024 Tecnativa - Pilar Vargas
+from odoo import fields, models
+
+
+class EventEvent(models.Model):
+    _inherit = "event.event"
+
+    allow_members_only = fields.Boolean(help="Allow registration only to members")
```
pedrobaeza:
```suggestion
    registration_membership_only = fields.Boolean(help="Allow registration only to members")
```

pedrobaeza (changes requested):

website_event_membership_restriction/models/event_event.py
pilarvargas-tecnativa: All changes done! thanks

website_event_membership_restriction/views/event_templates.xml
```diff
@@ -0,0 +1,61 @@
...
+        >
+            <attribute name="class">d-none</attribute>
+        </xpath>
+        <xpath
+            expr="//form//div[@id='o_wevent_tickets_collapse']//button"
+            position="after"
+        >
+            <button
+                type="submit"
+                t-att-class="'btn btn-primary o_wait_lazy_js a-submit ' + (event.registration_membership_only and (not request.env.user or request.env.user.partner_id.commercial_partner_id.membership_state not in ['paid', 'free'] or request.env.user.partner_id.membership_state not in ['paid', 'free']) and 'disabled' or '')"
```
chienandalu: Do this condition make sense? Should it be something like
```
event.registration_membership_only and not request.env.user._is_public() and (... or ...)

pilarvargas-tecnativa: Change made to check with ‘is_public()’.

This way logging will be disabled for the following cases:

- registration_membership_only = true and user not logged in.
- registration_membership_only = true and partner state not free, paid
- registration_membership_only = true and state of partner commercial not free, paid


thanks!

pedrobaeza (approved):
/ocabot merge nobump

PR #376, opened 2024-06-04
[16.0][IMP-FIX] website_event_membership_restriction: Use variable to condition the button classes

pilarvargas-tecnativa:
Use variable to condition the button classes to avoid errors in the views and also adds better readability.

cc @Tecnativa

@pedrobaeza @chienandalu please review

### DIFF
--- a/website_event_membership_restriction/views/event_templates.xml
+++ b/website_event_membership_restriction/views/event_templates.xml
@@ -7,6 +7,7 @@
         inherit_id="website_event.registration_template"
         priority="999"
     >
+        <!-- Hide the existing element and recreate it as both class and t-att-class can't coexist -->
         <xpath
             expr="//form//div[@id='o_wevent_tickets_collapse']//button"
             position="attributes"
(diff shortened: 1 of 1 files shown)


website_event_membership_restriction/views/event_templates.xml
```diff
@@ -17,9 +17,13 @@
             expr="//form//div[@id='o_wevent_tickets_collapse']//button"
             position="after"
         >
+            <t
+                t-set="is_disabled"
```
chienandalu: Better not use such a generic name, as it could collide with other extensions.

pilarvargas-tecnativa: changed

website_event_membership_restriction/views/event_templates.xml
```diff
@@ -17,9 +17,13 @@
             expr="//form//div[@id='o_wevent_tickets_collapse']//button"
             position="after"
         >
+            <t
+                t-set="is_registration_disabled"
```
chienandalu: This is still too generic... why not something like is_disabled_for_non_members

pilarvargas-tecnativa: thanks

chienandalu (approved):

pedrobaeza (approved):
/ocabot merge minor

PR #377, opened 2024-06-06
[16.0][FIX] website_event_membership_restriction: Wrong disabled condition

pedrobaeza:
The button should be disabled when both the own partner and its commercial partner aren't members, not one of them.

Duplicated variable evaluation removed as well.

@Tecnativa

### DIFF
--- a/website_event_membership_restriction/views/event_templates.xml
+++ b/website_event_membership_restriction/views/event_templates.xml
@@ -7,6 +7,15 @@
         inherit_id="website_event.registration_template"
         priority="999"
     >
+        <!-- Put this evaluation at the very beginning -->
+        <xpath expr="//div[@t-if='toast_message']" position="before">
+            <t t-set="user" t-value="request.env.user" />
+            <t t-set="partner" t-value="user.partner_id" />
+            <t
+                t-set="is_disabled_for_non_members"
+                t-value="event.registration_membership_only and (not user or (partner.commercial_partner_id.membership_state not in ['paid', 'free'] and partner.membership_state not in ['paid', 'free']))"
+            />
+        </xpath>
         <!-- Hide the existing element and recreate it as both class and t-att-class can't coexist -->
         <xpath
             expr="//form//div[@id='o_wevent_tickets_collapse']//button"
@@ -18,10 +27,6 @@
             expr="//form//div[@id='o_wevent_tickets_collapse']//button"
             position="after"
         >
-            <t
-                t-set="is_disabled_for_non_members"
-                t-value="event.registration_membership_only and (not request.env.user or request.env.user.partner_id.commercial_partner_id.membership_state not in ['paid', 'free'] or request.env.user.partner_id.membership_state not in ['paid', 'free'])"
-            />
             <button
                 type="submit"
                 t-att-class="'btn btn-primary o_wait_lazy_js a-submit ' + ('disabled' if is_disabled_for_non_members else '')"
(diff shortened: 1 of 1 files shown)


website_event_membership_restriction/views/event_templates.xml
```diff
@@ -43,9 +45,11 @@
             <attribute name="class">d-none</attribute>
         </xpath>
         <xpath expr="//form//div[@t-else='']//button" position="after">
+            <t t-set="user" t-value="request.env.user" />
+            <t t-set="partner" t-value="user.partner_id" />
             <t
                 t-set="is_disabled_for_non_members"
-                t-value="event.registration_membership_only and (not request.env.user or request.env.user.partner_id.commercial_partner_id.membership_state not in ['paid', 'free'] or request.env.user.partner_id.membership_state not in ['paid', 'free'])"
+                t-value="event.registration_membership_only and (not user or (partner.commercial_partner_id.membership_state not in ['paid', 'free'] and partner.membership_state not in ['paid', 'free']))"
```
chienandalu: Don't we have it already from the initial t-set?

pedrobaeza: True, removed the duplicated variable assignation.

chienandalu (approved):

website_event_membership_restriction/views/event_templates.xml
```diff
@@ -43,10 +45,6 @@
             <attribute name="class">d-none</attribute>
         </xpath>
         <xpath expr="//form//div[@t-else='']//button" position="after">
-            <t
-                t-set="is_disabled_for_non_members"
-                t-value="event.registration_membership_only and (not request.env.user or request.env.user.partner_id.commercial_partner_id.membership_state not in ['paid', 'free'] or request.env.user.partner_id.membership_state not in ['paid', 'free'])"
-            />
```
pilarvargas-tecnativa, line 49: If you remove the condition for the second button, the events using this button remain active because the condition is not defined for this button.
NOT member
image
Only members for this event:
image
Can be registered
image

website_event_membership_restriction/views/event_templates.xml
```diff
@@ -18,9 +18,11 @@
             expr="//form//div[@id='o_wevent_tickets_collapse']//button"
             position="after"
         >
+            <t t-set="user" t-value="request.env.user" />
+            <t t-set="partner" t-value="user.partner_id" />
             <t
                 t-set="is_disabled_for_non_members"
-                t-value="event.registration_membership_only and (not request.env.user or request.env.user.partner_id.commercial_partner_id.membership_state not in ['paid', 'free'] or request.env.user.partner_id.membership_state not in ['paid', 'free'])"
+                t-value="event.registration_membership_only and (not user or (partner.commercial_partner_id.membership_state not in ['paid', 'free'] and partner.membership_state not in ['paid', 'free']))"
```
pilarvargas-tecnativa: Can't it be the case that the commercial partner is a member and the partner is not, and vice versa?
```suggestion
                t-value="event.registration_membership_only and (not user or (partner.commercial_partner_id.membership_state not in ['paid', 'free'] or partner.membership_state not in ['paid', 'free']))"
```
???

pedrobaeza: To disable the button, both the partner and the commercial partner must be non member. What you are suggesting is again the same that it was, which was incorrect.

website_event_membership_restriction/views/event_templates.xml
pedrobaeza: I'm not removing the condition (as the button still has the `t-att-class`). I'm just removing to evaluate again the same variables that have been already evaluated before.

pedrobaeza: Moved to the beginning to be evaluated for all the cases.

pilarvargas-tecnativa (approved):
Thanks!

PR #381, opened 2024-07-04
[17.0][MIG] partner_event: Migration to version 17.0

juanjosesegui-tecnativa:
@Tecnativa TT49923


chienandalu (commented):
The field `mobile` has been deleted in the `event.registration`  model: https://github.com/OCA/OpenUpgrade/blob/9279df155c10dcf91333ba311375b1f06cb88c6c/openupgrade_scripts/scripts/event/17.0.1.8/upgrade_analysis.txt#L21

You'll have to adapt the code to that change. Options:

- Simply ignore it the partner.
- Fallback from one to another (although here you have to decide which one has precedence...)

juanjosesegui-tecnativa:
@Tecnativa TT49923

PR #382, opened 2024-07-05
[17.0][MIG] partner_event: Migration to version 17.0 

juanjosesegui-tecnativa:
@Tecnativa TT49923


partner_event/__manifest__.py
```diff
@@ -6,17 +6,18 @@
 # Copyright 2018 Jupical Technologies Pvt. Ltd. - Anil Kesariya
 # Copyright 2020 Tecnativa S.L. - Víctor Martínez
 # Copyright 2023 Tecnativa S.L. - Carolina Fernandez
+# Copyright 2024 Tecnativa S.L. - Juan José Seguí
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 {
     "name": "Link partner to events",
-    "version": "16.0.1.0.1",
+    "version": "17.0.0.0.1",
```
chienandalu: Check https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#version-numbers
```suggestion
    "version": "17.0.1.0.0",
```

partner_event/__manifest__.py
```diff
@@ -6,17 +6,18 @@
...
     "category": "Marketing",
     "author": "Tecnativa," "Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/event",
     "development_status": "Production/Stable",
     "license": "AGPL-3",
-    "depends": ["event"],
+    "depends": ["event", "website"],
```
chienandalu: Why pulling this dependency?

partner_event/models/event_registration.py
```diff
@@ -5,6 +5,7 @@
 # Copyright 2020 Tecnativa - Víctor Martínez
 # Copyright 2014-2023 Tecnativa - Pedro M. Baeza
 # Copyright 2023 Tecnativa - Carolina Fernandez
+# Copyright 2024 Tecnativa - Juan José Seguí
```
chienandalu, line 8: To change a tiny bit of the code doesn't entitle authorship. Put yourself in the contributors thou

chienandalu (changes requested):

partner_event/models/event_registration.py
pedrobaeza: This is not really change authorship, but adding the copyright. It can be accepted if there are enough changes in the file (and not only cosmetic).

pedrobaeza:
Check that there are 2 migration commits, while it should be one.

juanjosesegui-tecnativa:
I have fixed the errors in the tests, and the module is now correct.

ping @chienandalu 

@Tecnativa TT49923

partner_event/tests/test_event_registration.py
```diff
@@ -61,15 +62,24 @@ def test_create(self):
...
         self.registration_02.attendee_partner_id = self.partner_01
         self.registration_02.event_id = event_1
         self.partner_01.invalidate_recordset()
         self.assertEqual(self.partner_01.registration_count, 2)
+        self.registration_01.state = "cancel"
+        self.partner_01.invalidate_recordset()
+        self.assertEqual(self.partner_01.registration_count, 1)
+        self.registration_02.state = "done"
+        self.partner_01.invalidate_recordset()
+        self.assertEqual(self.partner_01.registration_count, 1)
```
chienandalu: Didn't the test work as it was?

juanjosesegui-tecnativa: No, the test as it was before threw the following error:

File "/opt/odoo/auto/addons/partner_event/tests/test_event_registration.py", line 64, in test_count_registrations
    self.assertEqual(self.partner_01.registration_count, 0)
AssertionError: 1 != 0

partner_event/models/res_partner.py
```diff
@@ -0,0 +1,46 @@
...
+    event_registration_ids = fields.One2many(
+        string="Event registrations",
+        comodel_name="event.registration",
+        inverse_name="attendee_partner_id",
+    )
+    registration_count = fields.Integer(
+        string="Attendances", compute="_compute_registration_count", store=False
+    )
+
+    @api.depends("event_registration_ids")
```
pedrobaeza: Instead of calling `invalidate_recordset()` in tests (which is somehow making tricks), you must add here in this depends the subfields (like `event_registration_ids.state`) for triggering the recompute automatically, as this is also going to provoke errors on its usage.

partner_event/tests/test_event_registration.py
chienandalu: Yes, the test is there to check that the module behaves in a certain way. If the tests don't work it could either be that the logic of the module has changed or that something in the core has changed and the tests need to be adapted. The latter is a stranger case and requires more argumentation :wink:

partner_event/models/res_partner.py
juanjosesegui-tecnativa: Done!

The tests pass perfectly with the change.

pedrobaeza: The change is not correct. You have to add both,  `event_registration_ids` and `event_registration_ids.state`, and any other field that can impact, and remove the `invalidate_recordset` calls in the tests.

juanjosesegui-tecnativa: Ok, I have applied the changes and everything works correctly.

pedrobaeza (approved):

partner_event/models/res_partner.py
```diff
@@ -0,0 +1,46 @@
...
+        inverse_name="attendee_partner_id",
+    )
+    registration_count = fields.Integer(
+        string="Attendances", compute="_compute_registration_count", store=False
+    )
+
+    @api.depends("event_registration_ids", "event_registration_ids.state")
+    def _compute_registration_count(self):
+        for partner in self:
+            partner.registration_count = len(
```
carlos-lopez-tecnativa: This is confusing for me. If the field is named `registration_count`, then if I register more than once, I expect the count to reflect that. Otherwise, the field name should be `event_count` instead of `registration_count`. what do you think?

image

image

partner_event/models/res_partner.py
```diff
@@ -0,0 +1,46 @@
...
+                self.env["event.registration"]
+                .search(
+                    [
+                        ("attendee_partner_id", "child_of", partner.id),
+                        ("state", "not in", ("cancel", "draft")),
+                    ]
+                )
+                .mapped("event_id")
```
carlos-lopez-tecnativa: Is possible use `read_group` instead of len(search().mapped())?

pedrobaeza: Current one offers the ORM caching system, and it can be more convenient for smaller numbers. `read_group` always need SQL for getting the numbers, so there are cases where it's worse.

partner_event/models/res_partner.py
chienandalu: Yes, is missleading. I tried to figure out the cause and it simply changed from counting the registrations to the mapped events. Let's change it just making a search_count

juanjosesegui-tecnativa: Captura de pantalla de 2024-07-18 09-20-08

But the field shows ‘Count of events…’ in the screenshots there are 2 different events and that’s why it shows 2.

I understand that the correct thing is to display 2.

chienandalu: It's missleading. You can have many attendances per event. When you click in the button you'll see a number of records that doesn't match the number printed on the button. This is an historic error, IMO

juanjosesegui-tecnativa: Captura de pantalla de 2024-07-18 09-48-14
I don’t know, those screenshots are from the contacts module, the events module correctly displays the number of attendees even if the same attendee covers several tickets, we would then have to change the ‘Count of events…’ to something else.

chienandalu: I'm not sure I understand. Simply change the method to count the number of attendees and the button tooltip text to reflect that

juanjosesegui-tecnativa: Done!

carlos-lopez-tecnativa: could you use `search_count` instead of `len(search)`

juanjosesegui-tecnativa: Done!

carlos-lopez-tecnativa (approved):
LGTM

chienandalu (approved):
Thanks :)

PR #388, opened 2024-07-09
[16.0][ADD] website_event_ticket_limit: new module

remi-filament:
This module allows to configure the maximum number of tickets that a user can buy at once.


pedrobaeza:
Why not taking the occasion to override the maximum 10 limit?

remi-filament:
> Why not taking the occasion to override the maximum 10 limit?

Hi @pedrobaeza thank you for looking into this. Good question, I guess you could have bad UX if one would register more than 10 tickets.
Maybe we could set number to max_ticket if different than 0 and keep 10 max if max_ticket = 0 (same default behaviour as Odoo one).

Let me know if you think this would be pertinent, I could update template to reflect that !

pedrobaeza:
OK, that sounds reasonable. Just document it in the README saying that this "limit" can serve to overpass standard 10 limit, putting a higher limit, hehe.

remi-filament:
Done

website_event_ticket_limit/models/event_ticket.py
```diff
@@ -0,0 +1,14 @@
...
+
+from odoo import fields, models
+
+
+class EventTicket(models.Model):
+    _inherit = "event.event.ticket"
+
+    max_ticket = fields.Integer(string="Max by order", default=0)
+
+    # ------------------------------------------------------
```
pedrobaeza: Please remove these comment lines that are useless (and I avoid them even if there are computed fields)

website_event_ticket_limit/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,3 @@
+This module allows you to configure the maximum number of tickets that a user can buy at once.
+
+This new field can also be used to raise Odoo current limit to 9 tickets that can be purchased from website.
```
pedrobaeza:
```suggestion
This new field can also be used to raise Odoo current limit below 9 tickets that can be purchased from website.
```

website_event_ticket_limit/models/event_ticket.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2023 Le Filament (https://le-filament.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import fields, models
+
+
+class EventTicket(models.Model):
+    _inherit = "event.event.ticket"
+
+    max_ticket = fields.Integer(string="Max by order", default=0)
```
pedrobaeza: Why having a separate field instead of reusing the existing one?

pedrobaeza (commented):
Don't know if the dropdown is added other way, or simply that its contents are bigger, but it appears with more width:

Selección_040

website_event_ticket_limit/models/event_ticket.py
remi-filament: Done

website_event_ticket_limit/readme/DESCRIPTION.rst
remi-filament: Done

website_event_ticket_limit/models/event_ticket.py
remi-filament: By existing one, I suppose you mean seats_max from event.type.ticket ?
It has not the same purpose : seats_max is used to limit the max number of this kind of tickets you can sell in total, when the new field limits the number of tickets you can sell per user.

For instance you may want to limit the max number of VIP tickets to 50 (because room does not allow for more, because you can only prepare food for up to 50 persons, or whatever reason), but each person can buy only 1 VIP tickets (and combined with other modules to limit duplicates or to authorize only logged in user you can enforce that no one can register twice or 2 tickets).

I understand from your question that the naming of the field was not clear enough so I renamed it `max_ticket_per_order`

remi-filament:
Thanks @pedrobaeza for your review. Regarding the size of select, it is because of the content I suppose (before you were not allowed to have more than 9, so that data inside selection always had only 1 figure length, now if you can get more than 9 you then have some selection with 2 figures inside so the selection is bigger...)

website_event_ticket_limit/models/event_ticket.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2023 Le Filament (https://le-filament.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import fields, models
+
+
+class EventTicket(models.Model):
+    _inherit = "event.event.ticket"
+
+    max_ticket_per_order = fields.Integer(string="Max by order", default=0)
```
pedrobaeza:
```suggestion
    max_ticket_per_order = fields.Integer(string="Max. per order", default=0)
```

website_event_ticket_limit/views/event_ticket_views.xml
```diff
@@ -0,0 +1,28 @@
...
+License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <!-- Tree view -->
+    <record id="event_event_ticket_view_tree_from_event" model="ir.ui.view">
+        <field name="name">event.event.ticket.view.tree.from.event</field>
+        <field name="model">event.event.ticket</field>
+        <field name="inherit_id" ref="event.event_event_ticket_view_tree_from_event" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='start_sale_datetime']" position="before">
+                <field name="max_ticket_per_order" optional="show" />
```
pedrobaeza, line 12: Please put it after the other maximum seats field.

website_event_ticket_limit/models/event_ticket.py
pedrobaeza: Yeah, I have mixed and although it's understandable to have a different one. The new name is better (although let's be consistent in the label - comments inline -)

pedrobaeza (approved):

remi-filament:
Thanks @pedrobaeza I have implemented your comments !

pedrobaeza:
Great, already approved.

pedrobaeza:
On the README there's still a reference to the old field label.

julienCozo (approved):

PR #394, opened 2024-07-25
[16.0][ADD] event_seat_reserve

imlopes:
Related PR's :

https://github.com/OCA/event/pull/395

https://github.com/OCA/event/pull/396


event_seat_reserve/models/event_event.py
```diff
@@ -0,0 +1,61 @@
...
+    seats_reserved_unconfirmed = fields.Integer(
+        string="Number of Reserved Attendees",
+        store=False,
+        readonly=True,
+        compute="_compute_seats",
+    )
+
+    @api.depends("seats_max", "registration_ids.state", "registration_ids.active")
+    def _compute_seats(self):
+        # Override the original method to add the new state and the compute for the new field
```
ivantodorovich, line 18: Ideally, let's not completely overwrite the original method.

Perhaps you can try something like this:

```
        res = super()._compute_seats()
        counts = dict()
        if self.ids:
            groups = self.env["event.registration"].read_group(
                [
                    ("event_id", "in", self.ids),
                    ("state", "=", "reserved"),
                ],
                groupby=["event_id"],
                fields=["event_id"],
            )
            counts = {g["event_id"][0]: g["event_id_count"] for g in groups}
        for rec in self:
            rec.seats_reserved_unconfirmed = counts.get(rec.id, 0)
            ... 
```

event_seat_reserve/models/event_registration.py
```diff
@@ -0,0 +1,20 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl-3.0).
+
+from odoo import fields, models
+
+
+class EventRegistration(models.Model):
+    _inherit = "event.registration"
+
+    state = fields.Selection(
+        selection_add=[("reserved", "Reserved")], ondelete={"reserved": "set default"}
```
ivantodorovich:
```suggestion
        selection_add=[("reserved", "Reserved"), ("open",)], ondelete={"reserved": "set default"}
```

to choose the position, so that its in between `draft` and `open`

event_seat_reserve/models/event_registration.py
```diff
@@ -0,0 +1,20 @@
...
+    )
+
+    def _need_pre_reservation(self):
+        return False
+
+    def action_set_reserved(self):
+        self.write({"state": "reserved"})
+        self.event_id._check_seats_availability()
+        self.event_ticket_id._check_seats_availability()
```
ivantodorovich: Better to do this in a `write` override, like it's done in core for the "confirmed" state

https://github.com/odoo/odoo/blob/52f188849ce3a9ad9fb4e065be6cbc3fecdefb45/addons/event/models/event_registration.py#L186-L195

This will make the "constraint" have a wider effect, not only when you reserve through `action_reserve`, but also if for some reason the state field is directly changed to `reserved`.

event_seat_reserve/models/event_ticket.py
```diff
@@ -0,0 +1,51 @@
...
+class EventTicket(models.Model):
+    _inherit = "event.event.ticket"
+
+    seats_reserved_unconfirmed = fields.Integer(
+        string="Number of Reserved Attendees", compute="_compute_seats"
+    )
+
+    @api.depends("seats_max", "registration_ids.state", "registration_ids.active")
+    def _compute_seats(self):
+        """Determine reserved, available, reserved but unconfirmed and used seats."""
```
ivantodorovich, line 15: Same comment as before: try to avoid the full overwrite

event_seat_reserve/models/event_registration.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class EventRegistration(models.Model):
+    _inherit = "event.registration"
+
+    state = fields.Selection(
+        selection_add=[("reserved", "Reserved")], ondelete={"reserved": "set default"}
+    )
+
+    def _need_pre_reservation(self):
```
ivantodorovich: I think this method is not needed, at least not in this module

event_sale_seat_reserve/models/sale_order.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl-3.0).
+
+from odoo import models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def action_draft(self):
```
ivantodorovich: Could you add unit tests for this flow?

event_sale_seat_reserve/tests/test_event_sale_seat_reserve.py
```diff
@@ -0,0 +1,95 @@
...
+        self.assertEqual(self.event_ticket.seats_reserved_unconfirmed, 1)
+        self.assertEqual(self.registration1.state, "reserved")
+        self.assertEqual(self.event.seats_available, 0)
+
+    def test_02_confirm_sale_no_available_seats(self):
+        """
+        Test that reserved quantity is set and if seats are available or not
+        """
+        self.sale_order1.action_confirm()
+        with self.assertRaises(ValidationError):
```
ivantodorovich: Use assertRaisesRegex and check the actual message, to ensure that's the exception you are looking for

event_sale_seat_reserve/tests/test_event_sale_seat_reserve.py
```diff
@@ -0,0 +1,95 @@
...
+            self.sale_order2.action_confirm()
```
ivantodorovich: Could you add more unit tests:

1. test that reservations are cancelled when the order is cancelled
2. test that you get new reservations when increasing the quantity on a confirmed order
3. test that you get some reservations cancelled when decreasing the quantity
4. ..plus any other flow you can think of

website_event_sale_seat_reserve/models/event_registration.py
```diff
@@ -0,0 +1,21 @@
...
+class EventRegistration(models.Model):
+    _inherit = "event.registration"
+
+    def _need_pre_reservation(self):
+        need_pre_reservation = super()._need_pre_reservation()
+
+        return need_pre_reservation and self.visitor_id
+
+    @api.model_create_multi
+    def create(self, vals_list):
```
ivantodorovich: need unit tests for an e-commerce flow

ivantodorovich (commented):
Thanks @imlopes !

I think it might be better for reviewers to split this PR in 3: one per module

event_sale_seat_reserve/models/sale_order.py
```diff
@@ -0,0 +1,14 @@
...
+        result = super().action_draft()
+        for so in self:
```
Ricardoalso: Here you can directly use `self.order_line._set_draft_associated_registrations()` instead of looping

event_sale_seat_reserve/models/sale_order_line.py
```diff
@@ -0,0 +1,38 @@
...
+            cancel_to_draft=cancel_to_draft,
+            registration_data=registration_data,
+            mark_as_paid=mark_as_paid,
+        )
+        RegistrationSudo = self.env["event.registration"].sudo()
+        registrations = RegistrationSudo.search(
+            [("sale_order_line_id", "in", self.ids), ("state", "=", "draft")]
+        )
+        for registration in registrations:
+            if registration._need_pre_reservation():
```
Ricardoalso: `def _need_pre_reservation(self):` always returns False in "event.registration" model (file event_seat_reserve/models/event_registration.py in this PR) or am I missing something ? :thinking:

event_seat_reserve/models/event_registration.py
imlopes: Ohh thanks ! I didn't know it could be done like this.

event_sale_seat_reserve/tests/test_event_sale_seat_reserve.py
imlopes: @ivantodorovich 
1. Normally it was done here : https://github.com/OCA/OCB/blob/94fa8d962535a754c9fb26997a38cc8b51b8974e/addons/event_sale/tests/test_event_sale.py#L473
The reservation is cancelled when the linked SO is cancelled too..

imlopes: For the point 2 and 3 I'm testing on the interface and we have no link between qty and reservations.
I mean, in standard, if we increase a qtty in a sale line, this won't create a new reservation. The same for decreasing..
IMHO, this can be another other OCA module. What do you think ?

ivantodorovich: > For the point 2 and 3 I'm testing on the interface and we have no link between qty and reservations.
I mean, in standard, if we increase a qtty in a sale line, this won't create a new reservation. The same for decreasing..

Oh, indeed. Let's not worry about that here then

event_seat_reserve/models/event_ticket.py
```diff
@@ -0,0 +1,41 @@
...
+        # initialize fields to 0 + compute seats availability
+        res = super()._compute_seats()
+        counts = dict()
```
ivantodorovich, line 18: This compute method is not covered by tests.
Could we add a test to check the `seats_reserved_unconfirmed` value?

ivantodorovich (approved):
LGTM :) thanks!!

event_seat_reserve/models/event_registration.py
```diff
@@ -0,0 +1,33 @@
...
+        selection_add=[("reserved", "Reserved"), ("open",)],
+        ondelete={"reserved": "set default"},
+    )
+
+    def action_set_reserved(self):
+        if config["test_enable"] and not self.env.context.get(
+            "test_event_seat_reserve"
+        ):
+            _logger.info("Test mode is enabled, you cannot reserve a registration.")
+            return True
```
ivantodorovich, line 24: Why do you need this?

ivantodorovich (changes requested):

event_seat_reserve/models/event_registration.py
imlopes: Some tests are failing on the two others PR's because when create a sale, normally the `event. registration` still on `draft` state but it's no more the case with these modules. So to avoid failing...

event_seat_reserve/models/event_event.py
```diff
@@ -0,0 +1,49 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl-3.0).
+from odoo import api, fields, models
+
+
+class EventEvent(models.Model):
+    _inherit = "event.event"
+
+    seats_reserved_unconfirmed = fields.Integer(
+        string="Number of Reserved Attendees",
```
StephaneMangin, line 10:
```suggestion
        string="Number of Unconfirmed Reserved Attendees",
```

StephaneMangin (approved):

StephaneMangin (approved):

ivantodorovich (approved):

cyrilmanuel:
superseed by https://github.com/OCA/event/pull/456

PR #395, opened 2024-07-31
[16.0][ADD] event_sale_seat_reserve

imlopes:
Depends on https://github.com/OCA/event/pull/394


event_sale_seat_reserve/__manifest__.py
```diff
@@ -0,0 +1,14 @@
...
+    "name": "Event Sale Seat Reserve",
+    "version": "16.0.1.0.0",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "website": "https://github.com/OCA/event",
+    "category": "Marketing",
+    "summary": "Allow pre-reserve registrations on sales",
+    "depends": ["event_sale", "event_seat_reserve"],
+    "installable": True,
+    "auto_install": False,
```
ivantodorovich:
```suggestion
    "auto_install": True,
```

As a glue module, I'd make this auto installable

ivantodorovich (approved):

event_sale_seat_reserve/models/sale_order_line.py
```diff
@@ -0,0 +1,45 @@
...
+        registrations.action_set_reserved()
+        return res
+
+    def _set_draft_associated_registrations(self):
+        RegistrationSudo = self.env["event.registration"].sudo()
+        registrations = RegistrationSudo.search(
+            [("sale_order_line_id", "in", self.ids), ("state", "!=", "done")]
+        )
+        registrations.action_set_draft()
+        return True
```
ivantodorovich, line 45: nitpicking: I'd narrow the scope to the reserved registrations only
and perhaps rename the method to reflect that

```suggestion
    def _unreserve_registrations(self):
        RegistrationSudo = self.env["event.registration"].sudo()
        registrations = RegistrationSudo.search(
            [("sale_order_line_id", "in", self.ids), ("state", "=", "reserved")]
        )
        registrations.action_set_draft()
        return True
```

ivantodorovich (approved):

event_sale_seat_reserve/models/sale_order_line.py
imlopes: This method is called from "Set to quotation" action.
Normally the button "Set to quotation" is displayed only after cancel the sale.
When cancelling a sale, all registrations linked to the sale lines will be cancelled too.

So here I'm trying to "re-activate" the linked registration : 

Sale order cancelled > sale order lines cancelled > registrations cancelled
Sale order set to draft > sale order lines set to draft > registrations set to reserved

I'm using `!= done` because it's possible that the user was changed manually the registration line to draft.

event_sale_seat_reserve/models/sale_order_line.py
```diff
@@ -0,0 +1,45 @@
...
+        )
+        RegistrationSudo = self.env["event.registration"].sudo()
+        registrations = RegistrationSudo.search(
+            [("sale_order_line_id", "in", self.ids), ("state", "=", "draft")]
+        )
+        if config["test_enable"] and not self.env.context.get(
+            "test_event_seat_reserve"
+        ):
+            _logger.info("Test mode is enabled, skipping the reservation of the seats")
+            return res
```
StephaneMangin, line 35: If you need to skip tests then your code is not properly testable. This is a big quality warning IMO

ivantodorovich: Yeap, I agree..

imlopes: I'm skipping test because in some standard tests, we're checking if a registration is on draft state.
But, with this module, a registration with a sale will be automatically on "reserved" state.

https://github.com/odoo/odoo/blob/4d9611924d753e314efdcc571c2bbe875b69e602/addons/event_sale/tests/test_event_sale.py#L286

pedrobaeza: This technique is totally legit to keep compatibility with old/other flows for the rest of the modules. It's not for this module. It's for not making the other tests (even standard ones) not to fail. I don't think there's other option.

ivantodorovich (approved):

cyrilmanuel:
superseeded here https://github.com/OCA/event/pull/457

PR #396, opened 2024-07-31
[16.0][ADD] website_event_sale_seat_reserve

imlopes:
Depends on:
https://github.com/OCA/event/pull/394
https://github.com/OCA/event/pull/395


website_event_sale_seat_reserve/models/event_registration.py
```diff
@@ -0,0 +1,21 @@
...
+
+        return need_pre_reservation and self.visitor_id
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        registrations = super(EventRegistration, self).create(vals_list)
+        for registration in registrations:
+            if registration._need_pre_reservation():
+                registration.action_set_reserved()
+        return registrations
```
ivantodorovich: For readability, I suggest something like this:

```py
    def _auto_reserve_registrations(self):
        to_reserve = self.filtered(
            lambda rec: (
                rec.state == 'draft'
                and rec.sale_order_line_id
                and rec.visitor_id
            )
        )
        to_reserve.action_set_reserved()

    @api.model_create_multi
    def create(self, vals_list):
        registrations = super().create(vals_list)
        registrations._auto_reserve_registrations()
        return registrations
```

Notice I added also a check for `state == 'draft'`.

As I recall, in core, when visitors register themselves on free events they get automatically confirmed registrations. So, just in case, it wouldn't hurt to check the state in case they are already confirmed

ivantodorovich (commented):
Unit tests for the e-commerce flow would be nice ^^

website_event_sale_seat_reserve/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+    "name": "Website Event Sale Seat Reserve",
+    "version": "16.0.1.0.0",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "website": "https://github.com/OCA/event",
+    "category": "Marketing",
+    "summary": "Allow pre-reserve registrations on website sales",
+    "depends": ["website_event_sale", "event_sale_seat_reserve"],
+    "installable": True,
+    "auto_install": False,
```
ivantodorovich:
```suggestion
    "auto_install": True,
```

Should probably be auto-installable, too :)

imlopes:
@ivantodorovich is failling because dependencies issues.
I've tried to put all together and it was OK: (https://github.com/OCA/event/pull/397)

test-requirements.txt
```diff
@@ -0,0 +1 @@
+odoo-addon-event_sale_seat_reserve @ git+https://github.com/OCA/event.git@refs/pull/395/head#subdirectory=event_sale_seat_reserve
```
ivantodorovich:
```suggestion
odoo-addon-event_sale_seat_reserve @ git+https://github.com/OCA/event.git@refs/pull/395/head#subdirectory=setup/event_sale_seat_reserve
```

test-requirements.txt
```diff
@@ -0,0 +1 @@
+odoo-addon-event_sale_seat_reserve @ git+https://github.com/OCA/event.git@refs/pull/395/head#subdirectory=setup/event_sale_seat_reserve
```
ivantodorovich, line 2: you need to add here the `event_seat_reserve` module dependency, too

imlopes:
@ivantodorovich do you know why unit tests are not running ?
I checked the log and no mention for..

ivantodorovich:
> @ivantodorovich do you know why unit tests are not running ? I checked the log and no mention for..

uhm, I don't know
if you run the same steps locally, does it work?

imlopes:
> > @ivantodorovich do you know why unit tests are not running ? I checked the log and no mention for..
> 
> uhm, I don't know if you run the same steps locally, does it work?

I've tried on local. I couldnt too :s

StephaneMangin (approved):

ivantodorovich (approved):

PR #407, opened 2024-11-05
[FIX] [16.0] website_event_contact: Allow portal users view contacts

Shide:
Added sudo to allow display contact information when portal users are on the event.

MT-7710 @moduon @yajo @rafaelbn @gelojr please review if you want :)


yajo (approved):

rafaelbn (approved):
Fast-track 

/ocabot merge patch

PR #408, opened 2024-11-05
[WIP][MIG] Migration and improvements of event_session module

Rom10811:
We plan to add a new method of creating sessions based on opening rules.


pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0.

If the jump is between several versions, you have to modify the source branch in the main command to accommodate it to this circumstance.

PR #409, opened 2024-11-14
[17.0][MIG] event_mail: Migration to 17.0

victoralmau:
Migration to 17.0

Please @sergio-teruel and @chienandalu can you review it?

@Tecnativa TT51530


sergio-teruel:
Where can I select de template??

image

victoralmau:
The field was shown (before the table) but without label, which caused confusion and made people think it didn't exist. I see that it is something that is like this (confusing) since v13 (or maybe before), I will solve it also in v16 at least.

**Before**
ejemplo

**After**
despues

chienandalu (approved):
Test and code review :+1:

sergio-teruel (approved):

PR #411, opened 2024-11-15
[17.0][MIG] website_event_membership_restriction: Migration to version 17.0

pilarvargas-tecnativa:
cc @Tecnativa TT51547

@chienandalu @carlos-lopez-tecnativa please review


chienandalu (commented):
I'm testing the functionality in runboat and it doesn't seem to work. I set up the membership of the portal user but I still can't register in the restricted event...

pilarvargas-tecnativa:
> I'm testing the functionality in runboat and it doesn't seem to work. I set up the membership of the portal user but I still can't register in the restricted event...

I have tested it and it works correctly, maybe due to the validity date of the membership it was not active?

image

image

chienandalu:
I could make it work, but now I cancelled the subscription but I can still register...

image

pilarvargas-tecnativa:
> I could make it work, but now I cancelled the subscription but I can still register...

Maybe it's a runboat problem?

image

pilarvargas-tecnativa:
Try two different users in runbot, DEMO has paid membership and ADMIN does not:
image

image

I now cancel the DEMO membership:
image

This problem arises in runbot, it is not updating the user's membership status. To update it you have to log out and log in again but this problem does not arise in local that the update of the membership status is correct without having to log out. In any case, the problem is not related to the module, it is a runbot issue.

I will leave in the template the data taken by screen in case you want to do more tests

pedrobaeza:
But this is something outside of the scope of the module AFAIK.

chienandalu (approved):
Ok, I'll take your word :sweat_smile:

carlos-lopez-tecnativa (changes requested):
Tested and confirmed working for me. Can this content be removed?

image

pilarvargas-tecnativa:
> Tested and confirmed working for me. Can this content be removed?
That content is in the runbot view, when the runbot is reset it will be deleted.

carlos-lopez-tecnativa (approved):

PR #414, opened 2024-12-03
[17.0][MIG] crm_event: Migration to version 17.0

pilarvargas-tecnativa:
cc @Tecnativa TT51918

@chienandalu @victoralmau please review


crm_event/models/event_type.py
```diff
@@ -0,0 +1,143 @@
...
+        for group in results:
+            event_type = group["event_type_id"][0]
+            events = self.env["event.event"].search(
+                [
+                    ("event_type_id", "=", event_type),
+                    ("date_end", ">=", fields.Date.today()),
+                ]
+            )
+            total_seats = sum(event.seats_available for event in events)
+            group["seats_available"] = total_seats
```
victoralmau: IMO this should not be necessary, the `seats_available` field is already being used in read_group and the domain as well.

crm_event/tests/test_event_type.py
```diff
@@ -0,0 +1,105 @@
+# Copyright 2021 Tecnativa - Jairo Llopis
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from datetime import datetime, timedelta
+
+from odoo.tests.common import TransactionCase
+
+
+class CrmEventCase(TransactionCase):
```
victoralmau:
```suggestion
from odoo.addons.base.tests.common import BaseCommon


class CrmEventCase(BaseCommon):
```

victoralmau (changes requested):

crm_event/models/event_type.py
pilarvargas-tecnativa: So the solution is to remove the `read_group` for the totals calculation because the `seats_available` field in v17 is not stored and operates directly on data stored in the database.

crm_event/tests/test_event_type.py
pilarvargas-tecnativa: thanks!

chienandalu (approved):
Tested and working

victoralmau (approved):

PR #415, opened 2024-12-09
[17.0][MIG] event_min_seat: Migration to version 17.0

pilarvargas-tecnativa:
cc @Tecnativa TT51923

@chienandalu @pedrobaeza please review


event_min_seat/tests/test_event_min_seat.py
```diff
@@ -19,7 +19,7 @@ def setUpClass(cls):
         cls.event = cls.env["event.event"].create(
             {
                 "name": "Test event",
-                "event_type_id": cls.env.ref("event.event_type_data_ticket").id,
+                "event_type_id": cls.env.ref("event.event_type_0").id,
```
pedrobaeza: Why using this demo data instead of the type created in the test (or create another one if needed)?

pedrobaeza (approved):

event_min_seat/tests/test_event_min_seat.py
pilarvargas-tecnativa: It was created with demo data, but I make the change to save this maintenance in future migrations.

pilarvargas-tecnativa: change done, thanks

chienandalu (approved):
Tested :+1:

PR #420, opened 2024-12-13
[17.0][FIX] website_event_filter_city: Country filter support

pilarvargas-tecnativa:
When the country filter is inactive, the node used is a span but when the country filter is activated with the editor, the node is replaced by a div and this causes an error. To avoid this, the node is replaced by something generic ‘*’ and so it will be located by its identifying class regardless of whether the element is a span or a div.

cc @tecnativa

@chienandalu @carlos-lopez-tecnativa  please review

### DIFF
--- a/website_event_filter_city/README.rst
+++ b/website_event_filter_city/README.rst
@@ -48,8 +48,8 @@ To configure this module, you need to:
 Known issues / Roadmap
 ======================
 
-- You cannot run the tests unless you are in a bare database with demo
-  data.
+-  You cannot run the tests unless you are in a bare database with demo
+   data.
 
 Bug Tracker
 ===========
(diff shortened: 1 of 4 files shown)


pedrobaeza:
Take the occasion to change the maintainer to us, as we are the effective maintainers of the module now.

carlos-lopez-tecnativa (approved):
It’s a bit strange; I can't reproduce the issue on any Runboat instance, but in this specific instance, the error is present. Thanks for the fix.

http://oca-event-17-0-030f8b6af6fc.runboat.odoo-community.org/ 

image

Note: I think it’s a bad practice from Odoo Base if the original view uses a `<span>` element and another view replaces it and changes it by `<div>`. Could you create an issue on the Odoo repository?

Original view
https://github.com/odoo/odoo/blob/f6fc5f889d041e471ec3af0fa9ec931e00923be3/addons/website_event/views/event_templates_list.xml#L120

Inherited view
https://github.com/odoo/odoo/blob/f6fc5f889d041e471ec3af0fa9ec931e00923be3/addons/website_event/views/event_templates_list.xml#L225-L237

pilarvargas-tecnativa:
> It’s a bit strange; I can't reproduce the issue on any Runboat instance, but in this specific instance, the error is present. 

Sometimes runboats are not entirely correct in terms of templates or styles, they can fail and I don't know why, but we always have the option of testing the branch locally.

chienandalu (approved):
thanks :)

PR #421, opened 2024-12-18
[17.0][MIG] event_sale_reservation: Migration to version 17.0

pilarvargas-tecnativa:
cc @tecnativa TT51924

@victoralmau @chienandalu please review


chienandalu (changes requested):
One behavior I detected is that when you register the reservation into an event, the ticket price overwrites the quotation one. I compared with the previous version and that didn't happen. Can you check it?

pilarvargas-tecnativa:
> One behavior I detected is that when you register the reservation into an event, the ticket price overwrites the quotation one. I compared with the previous version and that didn't happen. Can you check it?

By converting the reserve product into the registration product, the price of the registration product was being set. Forcing the data when creating the registration from the booking keeps the price. The correct behaviour is the one you define. Tested also with real data.

chienandalu (approved):
thanks :)

event_sale_reservation/wizards/registration_editor.py
```diff
@@ -0,0 +1,62 @@
...
+                    "event_id": line.event_id.id,
+                    "event_ticket_id": line.event_ticket_id.id,
+                    "product_id": line.event_ticket_id.product_id.id,
+                    "price_unit": line.sale_order_line_id.price_unit,
+                }
+            )
+        # Close current wizard and reopen normally to configure registrations
+        upstream_view = self.env.ref("event_sale.view_event_registration_editor_form")
+        return {
+            "type": "ir.actions.act_window",
```
victoralmau, line 51:
```suggestion
            "type": "ir.actions.client",
            "tag": "soft_reload",
```
According to https://github.com/OCA/OpenUpgrade/pull/4230 this should be the case, right (or maybe I'm forgetting something)?

event_sale_reservation/__manifest__.py
```diff
@@ -4,19 +4,17 @@
...
     "website": "https://github.com/OCA/event",
     "author": "Tecnativa, Odoo Community Association (OCA)",
-    "maintainers": ["Yajo"],
+    "maintainers": ["pilarvargas-tecnativa"],
     "license": "AGPL-3",
     "application": False,
     "installable": True,
     "depends": [
         "event_sale",
-        "web_ir_actions_act_multi",
```
victoralmau: `web_ir_actions_act_multi` is not necessary? IMO if the functionality that this module had is not necessary it would be necessary to define it in Openupgrade

event_sale_reservation/tests/test_event_sale.py
```diff
@@ -0,0 +1,162 @@
...
+# Copyright 2023 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from datetime import datetime, timedelta
+
+from odoo.tests.common import Form, TransactionCase
+
+from ..exceptions import ReservationWithoutEventTypeError
+
+
+class EventSaleCase(TransactionCase):
```
victoralmau:
```suggestion
from odoo.addons.base.tests.common import BaseCommon

from ..exceptions import ReservationWithoutEventTypeError


class EventSaleCase(BaseCommon):
```

victoralmau (changes requested):
Functional review OK.

event_sale_reservation/wizards/registration_editor.py
pilarvargas-tecnativa: If that is correct for when you need to reload the page in this case a new wizard is opening and you don't need to reload the page.

event_sale_reservation/__manifest__.py
pilarvargas-tecnativa: ir.actions.act_multi was being used to chain several actions that odoo didn't have, in this module close one wizard and open another. This is no longer necessary in this module as a wizard closes automatically when doing any interaction. That way when you open the next wizard you don't have to tell it to close it because odoo already closes it. In the same way, the method that returned a reload action has been removed because when the last wizard is closed, the order view is reloaded.

How do you think it should be reflected in OU?

victoralmau (approved):

PR #423, opened 2024-12-27
[17.0][MIG] event_project

rrebollo:
standard migration from v16 with several coding improvements:

- split business logic between event and project models
- sql constraint to avoid settle same project for different events (test added)
- refactor tests
- improve documentation
- replace custom project.task action by built-in
- add project search capabilities to event search view
- ensure groups event manager in views

@BinhexTeam


rrebollo:
@jelenapoblet Could you please do the functional review? I think the documentation could be improved. Maybe after testing it, you can come up with some specific suggestions.

adasatorres (approved):
LGTM

arielbarreiros96 (approved):
Migration LGTM

On the functional side however it would be nice if the propagation also worked backwards. I think (non-technical) users could be led to believe that the link works in both directions, but changing the fields in the projects will not affect the event, furthermore, if changes in the event are made after making them in the project, the changes in the project are overwritten. This is not a bug or an error, but an improvement that can be introduced, what do you think?

jelenapoblet:
@arielbarreiros96 your suggestion is a nice improvement, but not contemplated in this migration, I would merge this, and if there's an improvement proposition, it could be done in another pr

arielbarreiros96:
> @arielbarreiros96 your suggestion is a nice improvement, but not contemplated in this migration, I would merge this, and if there's an improvement proposition, it could be done in another pr

Indeed it looks more like a later improvement to be done, better to keep the migration clean

rrebollo:
@arielba1rreiros96 The addon has undergone significant refactoring in recent migrations, which has also led to a reduced scope. Implementing two-way synchronization would indeed be a valuable improvement, but it would be better addressed in a separate PR.

edescalona (approved):
LGTM

PR #428, opened 2025-01-29
[16.0][IMP] partner_event: use order in search by email

fernandahf:
The search by email is limited to 1 record and without the order clause is using the default order that is defined in https://github.com/odoo/odoo/blob/7f3863bdeb21ecfb7ac1859641bddf3de5dd8643/odoo/addons/base/models/res_partner.py#L178 and it could be customized. 

Since the default order could cause a slow query this adds an "order" clause in the search to prevent the use of the default order and use the ID instead.

For example, in this case where without the order clause is using display_name in the order by default:

Screenshot 2025-01-29 at 12 42 48 p m

The partner result of the search must match with the email search, so any result is useful for this case.

### DIFF
--- a/partner_event/models/event_registration.py
+++ b/partner_event/models/event_registration.py
@@ -39,7 +39,9 @@ class EventRegistration(models.Model):
             Event = self.env["event.event"]
             # Look for a partner with that email
             email = vals.get("email").replace("%", "").replace("_", "\\_")
-            attendee_partner = Partner.search([("email", "=ilike", email)], limit=1)
+            attendee_partner = Partner.search(
+                [("email", "=ilike", email)], limit=1, order="id"
+            )
             event = Event.browse()
             if vals.get("event_id"):
                 event = Event.browse(vals["event_id"])


moylop260:
LGTM 👍 

Could you please run `pre-commit run` locally to auto-fix the lint?

fernandahf:
@luisg123v @imanie383  

Could you review this, please?

fernandahf:
@moylop260 

The lint was fixed.

moylop260 (approved):
👍

luisg123v (changes requested):
I think commit description could be improved to express better what do you want to achieve and what do you mean by "the default order". If you order by ID, won't that add an `ORDER BY` clause anyway?

pedrobaeza:
But that default order may select a better 1 record than by ID, isn't it?

fernandahf:
@luisg123v 

The default order is based on this https://github.com/odoo/odoo/blob/7f3863bdeb21ecfb7ac1859641bddf3de5dd8643/odoo/addons/base/models/res_partner.py#L178

But if some of those fields are heavy, it's going to slow the search query, so, as the search is using a limited record set, it's better to have the order only by ID.

moylop260:
It looks good to me since that Odoo is sorting search with fields without index and it is slow for big tables

Using `order="id"` it is ensure that a indexed field is used

Also, using `order=None` should work but Odoo auto-add the default value if it doesn't have value

Checking the following discussion about:
 - https://github.com/odoo/odoo/pull/46392

luisg123v:
@fernandahf,

1. As commented by @pedrobaeza, be aware the result could be different, as the used order is different too. Could that be a problem?
2. I'm not against the change, but I think commit description should be inproved. If it weren't for your comments, it wouldn't fully get the point.

pedrobaeza:
Yes, it may impact in edge cases. Just document it and it's good to go, as I'm not opposing to it. It's just to highlight that the sentence "so the default order is unnecessary" is not true at all.

fernandahf:
@pedrobaeza @luisg123v 

The commit and MR description were updated.

pedrobaeza (approved):
/ocabot merge patch

pedrobaeza:
Please forward-port it to upper versions

PR #436, opened 2025-02-03
[18.0][MIG] event_sale_reservation: Migration to 18.0

BhaveshHeliconia:


BhaveshHeliconia:
Hey @OCA/crm-sales-marketing-maintainers would be great if someone could have a look at this module.

FrancoMaxime (approved):
LGTM: code review

BhaveshHeliconia:
Hey @OCA/e-learning-maintainers would be great if someone could have a look at this module.

event_sale_reservation/models/product_template.py
```diff
@@ -2,15 +2,15 @@
...
+from odoo import api, fields, models
 
 from ..exceptions import ReservationWithoutEventTypeError
 
 
 class ProductTemplate(models.Model):
     _inherit = "product.template"
 
-    detailed_type = fields.Selection(
+    type = fields.Selection(
```
etobella: Shouldn't be better to use service_tracking?

A migration script might be needed

etobella (commented):
Technical review

event_sale_reservation/models/product_template.py
BhaveshHeliconia: I’ve updated it as per your suggestion and used service_tracking. Thanks for your suggestion. Kindly please review.

BhaveshHeliconia:
ping @etobella

event_sale_reservation/models/product_template.py
```diff
@@ -23,22 +23,23 @@ class ProductTemplate(models.Model):
...
             if not one.event_reservation_type_id:
                 raise ReservationWithoutEventTypeError(
                     self.env._("You must indicate event type for %(name)s.")
                 )
+
+    @api.onchange("service_tracking")
+    def _onchange_service_tracking_event_reservation(self):
+        if self.service_tracking == "event_reservation":
+            self.type = "service"
+            self.invoice_policy = "order"
```
rousseldenis, line 45: IMHO, this conflicts with `_compute_invoice_policy()`. You should override that function instead for this field.

BhaveshHeliconia: Thank you for the review and for raising this point.

I took another look at the implementation, and I believe the current onchange approach is the correct one and does not conflict with _compute_invoice_policy().

My reasoning is as follows:

1. It follows the Odoo core implementation
In Odoo 18 (event_sale), _onchange_type_event sets self.invoice_policy = "order" inside an @api.onchange("service_tracking") for the event tracking type. This module follows the same pattern for event_reservation.

2. There is no runtime conflict with _compute_invoice_policy()
_compute_invoice_policy() depends only on type, not on service_tracking. Additionally, it only assigns a value when invoice_policy is not already set, so the value assigned by the onchange is preserved and is not overwritten.

3. The field is intentionally designed to be overridable
The field is defined as store=True and readonly=False. This is the pattern used by Odoo to provide a computed default while still allowing users or onchange methods to override the value when needed. Moving this logic into a compute override would remove that flexibility and unnecessarily diverge from the core implementation.

For these reasons, I believe it is best to keep the current implementation, as it remains consistent with the Odoo core approach.

pilarvargas-tecnativa (changes requested):
The change to use “service_tracking” does not require a separate commit, as it is part of the migration.

It’s also better to use OU’s “map_values” method for the migration script. Here’s an example: https://github.com/OCA/OpenUpgrade/pull/5902/changes#diff-de0a318e2af2c2b6959298491e7dfda232085fa6ffa0842d3c59e1aa25859530R14

BhaveshHeliconia:
> The change to use “service_tracking” does not require a separate commit, as it is part of the migration.
> 
> It’s also better to use OU’s “map_values” method for the migration script. Here’s an example: https://github.com/OCA/OpenUpgrade/pull/5902/changes#diff-de0a318e2af2c2b6959298491e7dfda232085fa6ffa0842d3c59e1aa25859530R14

Done! I have squashed. Thanks

pilarvargas-tecnativa (commented):
Thank you. The migration script is necessary, although there’s no need to add a version higher than 18.0.1 as it hasn’t been merged yet. Simply carry out the migration using `map_values`, as shown in the example I linked to in my other comment.

event_sale_reservation/__manifest__.py
```diff
@@ -14,6 +14,7 @@
     "application": False,
     "installable": True,
     "depends": [
+        "product",
```
pilarvargas-tecnativa: Is it necessary?

event_sale_reservation/models/sale_order.py
```diff
@@ -0,0 +1,51 @@
+# Copyright 2021 Tecnativa - Jairo Llopis
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import _, api, fields, models
```
pilarvargas-tecnativa:
```suggestion
from odoo import api, fields, models
```

event_sale_reservation/models/sale_order.py
```diff
@@ -0,0 +1,51 @@
...
+        for one in self:
+            one.event_registration_count = len(
+                one.mapped("order_line.event_registration_ids")
+            )
+
+    def action_open_event_registrations(self):
+        """Redirect user to event registrations related to this SO."""
+        return {
+            "domain": [("sale_order_id", "in", self.ids)],
+            "name": _("Attendees"),
```
pilarvargas-tecnativa:
```suggestion
            "name": self.env._("Attendees"),
```

pilarvargas-tecnativa (changes requested):

BhaveshHeliconia:
> Thank you. The migration script is necessary, although there’s no need to add a version higher than 18.0.1 as it hasn’t been merged yet. Simply carry out the migration using `map_values`, as shown in the example I linked to in my other comment.

Yes, the migration script is necessary. OpenUpgrade's core migration only handles standard detailed_type values. Since this module added a custom "event_reservation" value to detailed_type in 17.0, we need to explicitly copy and remap it to service_tracking for existing databases. Without it, any product with detailed_type = 'event_reservation' would lose its event reservation tracking after upgrade.

event_sale_reservation/migrations/18.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,20 @@
...
+from openupgradelib import openupgrade
+
+_columns_copy = {
+    "product_template": [("service_tracking", None, None)],
+}
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.copy_columns(env.cr, _columns_copy)
```
pilarvargas-tecnativa: I think this wouldn’t be necessary anymore, what do you think, @pedrobaeza ?
```suggestion

```

event_sale_reservation/models/product_template.py
pilarvargas-tecnativa: @rousseldenis Do you agree?

event_sale_reservation/migrations/18.0.1.0.0/pre-migration.py
pedrobaeza: Exactly, we don't need another copy of the column.

event_sale_reservation/migrations/18.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2026 Heliconia Solutions Pvt Ltd
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+_columns_copy = {
+    "product_template": [("service_tracking", None, None)],
+}
```
pilarvargas-tecnativa:
```suggestion

```

BhaveshHeliconia: Done. Thanks!

event_sale_reservation/readme/USAGE.md
```diff
@@ -0,0 +1,61 @@
...
+2.  There's a new smart button called *Reserved seats* with that count.
+3.  Click on it to get to the sales orders where the seats got reserved.
+
+But that counter will be probably zero when you install, so let's see
+how to increase it.
+
+To create an event reservation product:
+
+1.  Go to *Sales \> Products \> Products*.
+2.  Create one and set *Product Type* to *Event Reservation*.
```
pilarvargas-tecnativa: Here, you need to select “service” as the product type and choose “event booking” in the “create on order” field.

event_sale_reservation/models/sale_order.py
```diff
@@ -0,0 +1,51 @@
...
+
+    event_reservations_pending = fields.Integer(
+        compute="_compute_event_reservations_pending",
+        string="Pending event reservations",
+        help=(
+            "Indicates how many event reservations are still not linked to "
+            "any registration."
+        ),
+    )
+    event_registration_count = fields.Integer(
```
pilarvargas-tecnativa: Check whether Odoo already includes a field in the order for this purpose, i.e. one that performs exactly the same function.

Image

pilarvargas-tecnativa:
@eduezerouali-tecnativa please take a look

event_sale_reservation/models/sale_order.py
BhaveshHeliconia: Thanks @pilarvargas-tecnativa! You are right in Odoo 18, core `event_sale` natively includes `attendee_count` and the "Attendees" smart button on `sale.order`.

event_sale_reservation/views/sale_order_view.xml
```diff
@@ -17,17 +17,7 @@
...
-                    type="object"
-                    class="oe_stat_button"
-                    icon="fa-users"
-                    invisible="event_registration_count == 0"
-                >
-                    <field name="event_registration_count" widget="statinfo" />
-                </button>
-            </div>
+
             <xpath
```
pilarvargas-tecnativa, line 20: Please remove the blank line

pilarvargas-tecnativa (approved):
Just a few quick comments, reduce administrative commits, as far as I’m concerned, it looks fine. Thanks!

pedrobaeza (approved):
As the explanations are reasonable (coherence with core, although it's true that records created by code can have inconsistent values), let's move with this one for doing the migration to 19, and check there if it has been changed to computed. 

/ocabot merge nobump

PR #444, opened 2025-02-28
[17.0][FIX] partner_event: handle contacts with mobile field

em230418:
In Odoo event.registration.mobile field does not exist.

As result before this commit, if you manually create registration with partner that has "mobile" field set, KeyError will be raised.

Notes for future migrations to 19.0+ included.

### DIFF
--- a/partner_event/models/event_registration.py
+++ b/partner_event/models/event_registration.py
@@ -85,9 +85,16 @@ class EventRegistration(models.Model):
             self = self.with_context(**get_attendee_partner_address)
             for registration in self:
                 if registration.partner_id:
-                    registration.update(
-                        registration._synchronize_partner_values(
-                            registration.partner_id
-                        )
+                    vals = registration._synchronize_partner_values(
+                        registration.partner_id
                     )
+                    # On Odoo 19.0 "mobile" field is removed
+                    # https://github.com/odoo/odoo/pull/189739
+                    # so this should not be ported to 19.0+
+                    if "mobile" not in registration._fields and "mobile" in vals:
+                        mobile = vals.pop("mobile")
+                        phone = vals.get("phone") or mobile
+                        if phone:
+                            vals["phone"] = phone
+                    registration.update(vals)
         return {}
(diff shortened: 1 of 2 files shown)


em230418:
> [codecov/patch](https://github.com/OCA/event/pull/444/checks?check_run_id=37979069530) — 88.88% of diff hit (target 96.89%)

I don't know, what it asks from me

partner_event/models/event_registration.py
```diff
@@ -85,9 +85,18 @@ def _onchange_partner_id(self):
...
             for registration in self:
                 if registration.partner_id:
-                    registration.update(
-                        registration._synchronize_partner_values(
-                            registration.partner_id
-                        )
+                    vals = registration._synchronize_partner_values(
+                        registration.partner_id
                     )
+
```
pedrobaeza: Don't add empty lines inside method, and more if the previous style was this way.

partner_event/models/event_registration.py
```diff
@@ -85,9 +85,18 @@ def _onchange_partner_id(self):
...
+                    # On Odoo 19.0 "mobile" field is removed
+                    # https://github.com/odoo/odoo/pull/189739
+                    # so this should not be ported to 19.0+
+                    if "mobile" not in registration._fields and "mobile" in vals:
+                        mobile = vals.pop("mobile")
+                        phone = vals.get("phone") or mobile
+                        if phone:
+                            vals["phone"] = phone
+
```
pedrobaeza:
```suggestion
```

pedrobaeza (commented):
The codecov check is saying that the test coverage is decreasing, but it's not enforced. I see you have added regression tests, so maybe not all the new lines are covered by these tests.

pedrobaeza (approved):
Please squash both commits into one.

pedrobaeza:
@em230418 can you fw-port it to upper versions?

em230418:
> @em230418 can you fw-port it to upper versions?

Yes, I will. I hope next week.

em230418:
> @em230418 can you fw-port it to upper versions?

Looks like partner_event is not ported to 18.0 yet, so the only thing I did - mentioned this PR in 18.0 migration PR

PR #447, opened 2025-04-07
[16.0][FIX] event_session: ux_view on event_type

PlantBasedStudio:
The event model view was broken in its layout. An empty checkbox appears in the right-hand column, followed by the autoconfirmation text, which is not aligned with its own checkbox.

Before : 
image

After : 
image(4)


chienandalu (approved):
:+1:

pedrobaeza:
Please follow commit message guidelines including the module name:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

PR #449, opened 2025-04-22
[17.0][MIG] website_event_crm_invitation: Migration to version 17.0

pilarvargas-tecnativa:
cc @Tecnativa 

@carlos-lopez-tecnativa @juancarlosonate-tecnativa please review


website_event_crm_invitation/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    "development_status": "Production/Stable",
+    "category": "Marketing",
+    "website": "https://github.com/OCA/event",
+    "author": "Tecnativa, Odoo Community Association (OCA)",
+    "maintainers": ["pilarvargas-tecnativa"],
+    "license": "AGPL-3",
+    "installable": True,
+    "depends": ["crm_event", "website_event"],
+    "data": [
+        # "data/ir_cron_data.xml",
```
carlos-lopez-tecnativa: Why the cron is not loaded?

website_event_crm_invitation/data/mail_template_data.xml
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2021 Tecnativa - Jairo Llopis
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo noupdate="1">
+    <record id="crm_lead_event_type_tpl" model="mail.template">
+        <field name="name">Invite to event category</field>
+        <field name="auto_delete" eval="False" />
+        <field name="email_to">{{object.email_from or ''}}</field>
+        <field name="partner_to">{{object.partner_id.id or ''}}</field>
```
carlos-lopez-tecnativa, line 9: Please add a `description`. Without this field set, the template is hidden by default by the filter `Base Templates`
https://github.com/odoo/odoo/blob/5fdef9019db65749b6f1f80d828fda6bc2280f55/addons/mail/models/mail_template.py#L124-L129

image

carlos-lopez-tecnativa (changes requested):

website_event_crm_invitation/__manifest__.py
pilarvargas-tecnativa: Yes, it's just that not all changes had been saved in the commit. The commit is now complete

website_event_crm_invitation/data/mail_template_data.xml
pilarvargas-tecnativa: done

website_event_crm_invitation/models/event_type.py
```diff
@@ -0,0 +1,38 @@
...
+            ("seats_limited", "=", False),
+        ]
+        # Search limited events with available seats (not storable)
+        events_with_seats = self.env["event.event"].search(
+            [
+                ("seats_limited", "=", True),
+                ("date_begin", ">", fields.Datetime.now()),
+                ("website_published", "=", True),
+            ]
+        )
```
carlos-lopez-tecnativa: Please consider the base domain here too. Otherwise, it retrieves events from any event type.
I created a new event type without any events, and I got events from other event types, which I think is incorrect.

```suggestion
        base_domain = self._events_domain()
        base_domain += [
            ("date_begin", ">", fields.Datetime.now()),
            ("website_published", "=", True),
        ]
        domain = base_domain + [
            ("seats_limited", "=", False),
        ]
        # Search limited events with available seats (not storable)
        events_with_seats = self.env["event.event"].search(
            base_domain + [
                ("seats_limited", "=", True),
            ]
        )
```

carlos-lopez-tecnativa (changes requested):

website_event_crm_invitation/models/event_type.py
pilarvargas-tecnativa: done, thanks

carlos-lopez-tecnativa (approved):

website_event_crm_invitation/models/crm_lead.py
```diff
@@ -0,0 +1,176 @@
...
+        # Leads with falsy company won't get invitations. It's not usual anyway
+        for company in leads.company_id:
+            company_leads = leads.filtered(
+                lambda lead, company=company: lead.company_id == company
+            )
+            company_published_events = published_events.filtered(
+                lambda e, company=company: e.company_id == company
+            )
+            if not company_published_events.event_type_id:
+                continue
```
carlos-lopez-tecnativa, line 81: Is it possible to discard the notification if a previous message was already sent?

image

pilarvargas-tecnativa: It involves a change in behaviour that may not be the desired behaviour, and can also be configured to move up a stage. This behaviour should be continued if the purpose is to remember day by day the events of a category.

juancarlosonate-tecnativa (approved):

PR #451, opened 2025-05-07
[17.0][MIG] event_email_reminder: Migration to version 17.0

pilarvargas-tecnativa:
cc @Tecnativa TT51922

@victoralmau @carlos-lopez-tecnativa please review


event_email_reminder/tests/test_event_event.py
```diff
@@ -0,0 +1,135 @@
...
+    def setUpClass(cls):
+        super().setUpClass()
+        today = datetime.now()
+        date_event_1 = today + timedelta(days=7)
+        ctx = {
+            "mail_create_nolog": True,
+            "mail_create_nosubscribe": True,
+            "mail_notrack": True,
+            "no_reset_password": True,
+        }
```
victoralmau:
```suggestion
from odoo.addons.base.tests.common import BaseCommon

class TestEventEmailReminder(BaseCommon):
    @classmethod
    def setUpClass(cls):
        super().setUpClass()
        today = datetime.now()
        date_event_1 = today + timedelta(days=7)```

**Review by victoralmau — APPROVED** [review]
Code review OK

File: event_email_reminder/data/mail_template_data.xml [hunk 2]
```diff
@@ -0,0 +1,45 @@
...
+                        <td>Date Begin</td>
+                        <td>Event Name</td>
+                        <td>Date End</td>
+                        <td>State</td>
+                    </tr>
+                </thead>
+                <tr t-foreach="ctx['events']" t-as="event">
+                    <td><span t-out="event.date_begin" /></td>
+                    <td><span t-out="event.name" /></td>
+                    <td><span t-out="event.date_end" /></td>
```
carlos-lopez-tecnativa: I think this date should be rendered according to the user's time zone.

image

image

image

carlos-lopez-tecnativa (changes requested):

event_email_reminder/data/mail_template_data.xml
pilarvargas-tecnativa: From the client or from the event? which is how it is shown on the website.

event_email_reminder/tests/test_event_event.py
pilarvargas-tecnativa: done

event_email_reminder/data/mail_template_data.xml
carlos-lopez-tecnativa: > From the client or from the event? which is how it is shown on the website.

Based on your last comment and the fact that Odoo takes the timezone from the event, I think the timezone should also be taken from the event in this case, in order to preserve the website's behavior.
image

event_email_reminder/models/event_event.py
```diff
@@ -0,0 +1,77 @@
...
+                template = self.env.ref(
+                    "event_email_reminder.event_email_reminder_template"
+                )
+            else:
+                template = self.env["mail.template"].browse(template_id)
+            if partner_ids:
+                self._send_event_template(events, template, partner_ids)
+            else:
+                for user in events.mapped("user_id"):
+                    events = events.filtered(lambda x, user=user: x.user_id == user)
```
carlos-lopez-tecnativa: I think the email should only be sent if there are events.

This is Ok
image

This email should not be sent.

image

carlos-lopez-tecnativa (changes requested):

event_email_reminder/models/event_event.py
pilarvargas-tecnativa: done, thanks!

event_email_reminder/data/mail_template_data.xml
pilarvargas-tecnativa: done

carlos-lopez-tecnativa (approved):
LGTM
Non-blocking comment: It would be nice to add some margin or padding between the columns, they look a bit too close together.
Alternatively, adding borders could also help differentiate the columns.

image

PR #455, opened 2025-06-19
[16.0][ADD] event_ticket_order: new module

julienCozo:
New module to add ticket order in an event.


remi-filament (approved):
Thanks @julienCozo works great !

PR #456, opened 2025-06-24
[16.0][ADD] event_seat_reserve module

cyrilmanuel:
This module allows to make a pre-reserve on registrations.

Superseed of [394](https://github.com/OCA/event/pull/394)


event_seat_reserve/models/event_event.py
```diff
@@ -0,0 +1,49 @@
...
+from odoo import api, fields, models
+
+
+class EventEvent(models.Model):
+    _inherit = "event.event"
+
+    seats_reserved_unconfirmed = fields.Integer(
+        string="Number of Reserved Attendees",
+        store=False,
+        readonly=True,
```
yankinmax, line 12: if it's not stored and readonly - then classical compute, no need for these field attributes

event_seat_reserve/models/event_event.py
```diff
@@ -0,0 +1,49 @@
...
+        if self.ids:
+            groups = self.env["event.registration"].read_group(
+                [
+                    ("event_id", "in", self.ids),
+                    ("state", "=", "reserved"),
+                ],
+                groupby=["event_id"],
+                fields=["event_id"],
+            )
+            counts = {g["event_id"][0]: g["event_id_count"] for g in groups}
```
yankinmax, line 31:
```suggestion
register_data = self.env["event.registration"].with_context(active_test=False)._read_group(
    [
        ("event_id", "in", self.ids),
        ("state", "=", "reserved"),
    ],
    ["event_id"],
    ["event_id"]
)
mapped_data = dict((data['event_id'][0], data['event_id_count']) for data in register_data)
```
Such counts are more less standardized in Odoo

event_seat_reserve/models/event_registration.py
```diff
@@ -0,0 +1,33 @@
...
+from odoo.tools import config
+
+_logger = logging.getLogger(__name__)
+
+
+class EventRegistration(models.Model):
+    _inherit = "event.registration"
+
+    state = fields.Selection(
+        selection_add=[("reserved", "Reserved"), ("open",)],
```
yankinmax, line 15: why do we have a tuple without a value for a selection:
`("open", "Open")`

event_seat_reserve/models/event_ticket.py
```diff
@@ -0,0 +1,41 @@
...
+    )
+
+    @api.depends("seats_max", "registration_ids.state", "registration_ids.active")
+    def _compute_seats(self):
+        """Determine reserved, available, reserved but unconfirmed and used seats."""
+        # initialize fields to 0 + compute seats availability
+        res = super()._compute_seats()
+        counts = dict()
+        if self.ids:
+            groups = self.env["event.registration"].read_group(
```
yankinmax, line 20: same as above

event_seat_reserve/models/event_registration.py
gurneyalex: this is a selection_add, IIRC it means "reserved" should be before "open" in the list

yankinmax: TIL: we can force an index of the selection file values

PR #457, opened 2025-06-24
16.0[ADD] new module event sale seat reserve

cyrilmanuel:
This module allows to make a pre-reserve on registrations after confirming the sale order.

This PR is superseed of https://github.com/OCA/event/pull/395 and depend of https://github.com/OCA/event/pull/456


event_sale_seat_reserve/models/sale_order_line.py
```diff
@@ -0,0 +1,45 @@
...
+    _inherit = "sale.order.line"
+
+    def _update_registrations(
+        self,
+        confirm=True,
+        cancel_to_draft=False,
+        registration_data=None,
+        mark_as_paid=False,
+    ):
+        res = super(SaleOrderLine, self)._update_registrations(
```
yankinmax, line 21: is this tested?

PR #460, opened 2025-07-17
[15.0][FIX] sale_crm_event_reservation: Validate partner before creating order with Form to respect blocks

pilarvargas-tecnativa:
@Tecnativa 

Manage partner blocking when creating orders with Form, redirecting to standard action if there is no partner.
Order creation with Form is maintained and partner blocking is handled with conditional redirection to ensure that order lines are created correctly and all associated logic is executed, which does not happen if the standard action `action_new_quotation` is used, which only creates the order without lines.
However, when using `Form`, if the contact has a blocking message that prevents the partner from being assigned, order creation fails with a required field error. To avoid this, a conditional is added that detects if the partner could not be assigned in the form, and in that case, it redirects to the standard action `action_new_quotation`, passing the partner via context (`default_partner_id`) so that the blocking message is displayed normally, respecting the Odoo flow.
This achieves the following:
- Create the order with lines correctly when possible.
- Display the partner block if necessary without throwing errors.
- Do not lose the basic Odoo validation.

image

@pedrobaeza @victoralmau please review

### DIFF
--- a/sale_crm_event_reservation/wizards/crm_lead_event_sale_wizard.py
+++ b/sale_crm_event_reservation/wizards/crm_lead_event_sale_wizard.py
@@ -100,11 +100,20 @@ class CRMLeadEventSale(models.TransientModel):
         # Creating a sale order properly involves lots of onchanges, so here it
         # is better to use `Form` to make sure we forget none
         so_form = Form(self.env["sale.order"])
+        so_form.partner_id = self.opportunity_id.partner_id
+        # If the partner has configured a warning that blocks, the partner won't be
+        # assigned here, so let's redirects to the standard action of creating a quote
+        # from the opportunity, passing the `default_partner_id` context to follow the
+        # Odoo flow, thus displaying the block message normally and avoiding errors in
+        # creation.
+        if not so_form.partner_id:
+            return self.opportunity_id.with_context(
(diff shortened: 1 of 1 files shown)


sale_crm_event_reservation/wizards/crm_lead_event_sale_wizard.py
```diff
@@ -100,11 +100,19 @@ def action_generate(self):
         # Creating a sale order properly involves lots of onchanges, so here it
         # is better to use `Form` to make sure we forget none
         so_form = Form(self.env["sale.order"])
+        so_form.partner_id = self.opportunity_id.partner_id
+        # If there is no partner assigned in the form due to a block, it redirects to
```
pedrobaeza:
```suggestion
        # If the partner has configured a warning that blocks, the partner won't be assigned here, so let's redirects to
```

pedrobaeza (commented):
.

sale_crm_event_reservation/wizards/crm_lead_event_sale_wizard.py
pilarvargas-tecnativa: done, thanks

pedrobaeza (approved):

victoralmau (approved):

PR #425, opened 2025-01-14
[17.0][MIG] sale_crm_event_reservation: Migration to version 17.0

pilarvargas-tecnativa:
cc @Tecnativa TT51941

@chienandalu @carlos-lopez-tecnativa please review

PR #465, opened 2025-08-05
[17.0][FIX] sale_crm_event_reservation: Compare datetime properly

pedrobaeza:
Don't compare a date with a datetime field:

```
File "/opt/odoo/auto/addons/sale_crm_event_reservation/wizards/crm_lead_event_sale_wizard.py", line 96, in <lambda>
  or ticket.end_sale_datetime >= fields.Date.context_today(self)
TypeError: can't compare datetime.datetime to datetime.date
```

@Tecnativa TT57560

### DIFF
--- a/sale_crm_event_reservation/wizards/crm_lead_event_sale_wizard.py
+++ b/sale_crm_event_reservation/wizards/crm_lead_event_sale_wizard.py
@@ -93,7 +93,7 @@ class CRMLeadEventSale(models.TransientModel):
                 tickets = record.event_id.event_ticket_ids.filtered(
                     lambda ticket, record=record: (
                         not ticket.end_sale_datetime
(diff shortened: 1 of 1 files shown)


carlos-lopez-tecnativa (approved):
LGTM.
I see this comes from the migration related to my comment: https://github.com/OCA/event/pull/425#discussion_r1915347686.
I didn’t take into account that the field is a Datetime.

However, I think with the domain used in the search, it wouldn’t fail.
Now it fails because it is used in the code for comparison, not in search, due to the changes made in https://github.com/OCA/event/compare/14705156607245a1579bbc5860ddf4449a861e0b..2bdd3981056897034836f27a0f35902d7e5d54a1.

Anyway, thanks for the fix!
I’ll keep it in mind for other places, making sure to check whether a field is a Date or Datetime.

rvalyi (approved):

PR #435, opened 2025-02-03
[18.0][MIG] partner_event: Migration to 18.0

BhaveshHeliconia:

PR #467, opened 2025-08-19
[18.0][MIG] partner_event: Migration to 18.0

victoralmau:
Superseed https://github.com/OCA/event/pull/435

Migration to 18.0

Please @pedrobaeza and @carlos-lopez-tecnativa can you review it?

@Tecnativa TT57319


chienandalu (approved):
Thanks :wink: :+1:

partner_event/models/event_registration.py
```diff
@@ -40,9 +40,7 @@ def _update_attendee_partner_id(self, vals):
             Event = self.env["event.event"]
             # Look for a partner with that email
             email = vals.get("email").replace("%", "").replace("_", "\\_")
-            attendee_partner = Partner.search(
-                [("email", "=ilike", email)], limit=1, order="id"
-            )
+            attendee_partner = Partner.search([("email", "=ilike", email)], limit=1)
```
pedrobaeza: Why removing the `order` part? That was done for avoiding extra queries for sorting the results.

partner_event/tests/test_event_registration.py
```diff
@@ -120,13 +121,10 @@ def test_data_update(self):
         self.partner_01.write({"email": "new@test.com"})
         self.assertEqual(event_2.registration_ids.email, "new@test.com")
 
+    @mute_logger("odoo.sql_db", "odoo.models.unlink")
```
pedrobaeza, line 126: Why changing this if not benefit?

partner_event/tests/test_event_registration.py
```diff
@@ -157,3 +155,65 @@ def test_partner_mobile_and_phone(self):
             self.partner_with_phone_and_mobile.mobile,
             "Incorrect test. Partners phone and mobile must differ",
         )
+
+    def test_update_attendee_partner_id(self):
```
pedrobaeza: Are the new tests needed or they can be merged with existing ones for the missing covered parts?

partner_event/tests/test_event_registration.py
victoralmau: I added the decorator `@mute_logger(“odoo.models.unlink”)` because I don't like seeing logs that don't contribute anything, that is, to avoid something like this:

`odoo.models.unlink: User #1 deleted res.partner records with IDs: [60]`

Taking advantage of the fact that `mute_logger` was added, I added `odoo. sql_db` to remove it from https://github.com/OCA/event/blob/910806d43dd5a5be3b50d39c3202c1cc2fb25e16/partner_event/tests/test_event_registration.py#L126 and make everything more readable (or at least more readable for me).

pedrobaeza: Yeah, but the mute_logger was already added in the unlink operation. That's why I'm saying this.

partner_event/hooks.py
```diff
@@ -0,0 +1,17 @@
...
+
+def post_init_hook(env):
+    """Preload proper attendee partner for existing registrations using
+    the same rules the module does"""
+    attendees_emails = env["event.registration"].read_group(
+        [("email", "!=", False)], ["email"], groupby="email"
+    )
+    for email in attendees_emails:
+        attendee_partner = env["res.partner"].search(
+            [("email", "=ilike", email["email"])], limit=1
```
pedrobaeza: Put here `order` as well, and you can put on both searches the reasoning as comment for avoiding removing it more times.

partner_event/tests/test_event_registration.py
victoralmau: I have removed unnecessary tests, improved others, and added the “_appropriate_” test for `base.partner.merge`.

partner_event/tests/test_event_registration.py
victoralmau: Perhaps I am misunderstanding you, but both are necessary.

`odoo.sql_db`: Necessary if I want to remove it from https://github.com/OCA/event/blob/910806d43dd5a5be3b50d39c3202c1cc2fb25e16/partner_event/tests/test_event_registration.py#L126

`odoo.models.unlink`: Necessary to delete the log in https://github.com/OCA/event/blob/910806d43dd5a5be3b50d39c3202c1cc2fb25e16/partner_event/tests/test_event_registration.py#L131

pedrobaeza: OK, I haven't dug enough, but if you are sure, no more worries.

pedrobaeza (approved):
/ocabot merge nobump

PR #402, opened 2024-08-14
[16.0][ADD] website_event_external

remi-filament:
New module allowing to list external event on Odoo website and redirect user to external link for registration.

PR #470, opened 2025-08-20
[18.0][ADD]event_external_registration: external link for event registration

matteotognini:
This module allows you to use an external link for event registration. It’s useful when you want to keep Odoo’s event pages—with detailed event information—while event registration on a third-party site. When an external link is set, the standard Odoo registration modal is replaced by a button that opens the external page


hbrunn:
this looks like a less feature rich version of https://github.com/OCA/event/pull/402. better help get this merged and then migrate it to v18

PR #462, opened 2025-08-01
[18.0][MIG] event_registration_multi_qty: Migration to 18.0

victoralmau:
Migration to 18.0

The `seats_unconfirmed` field disappears in v17, nothing is mentioned about this in [OpenUpgrade](https://github.com/OCA/OpenUpgrade/blob/17.0/openupgrade_scripts/scripts/event/17.0.1.8/upgrade_analysis_work.txt), and there does not seem to be any field that actually replaces it.: https://github.com/odoo/odoo/blob/decfd154a0debc15d2cff2db0d67df18837bf06f/addons/event/models/event_event.py#L242 vs https://github.com/odoo/odoo/blob/b20aa39b71635bb071ced480fb80dfa8f464dfb5/addons/event/models/event_event.py#L320

Please @carlos-lopez-tecnativa and @pilarvargas-tecnativa can you review it?

@Tecnativa TT57528

PR #473, opened 2025-08-29
[18.0][MIG] event_sale_registration_multi_qty: Migration to 18.0

victoralmau:
Migration to 18.0

Locked by:
- [x] `event_registration_multi_qty`: https://github.com/OCA/event/pull/462

Please @carlos-lopez-tecnativa and @pilarvargas-tecnativa can you review it?

@Tecnativa TT57529


event_sale_registration_multi_qty/models/sale_order.py
```diff
@@ -0,0 +1,62 @@
...
+        )
+        attendee_count_data = {}
+        for registration in multi_qty_regs:
+            attendee_count_data.setdefault(registration["sale_order_id"][0], 0)
+            # We minorate 1 as it's already count in super() by every registration
+            attendee_count_data[registration["sale_order_id"][0]] += (
+                registration["qty"] - 1
+            )
+        for sale_order in self:
+            sale_order.attendee_count += attendee_count_data.get(sale_order.id, 0)
```
carlos-lopez-tecnativa: I'm not sure if this code can be updated to use the `_read_group` function to compute the sum in a single call. Could you please review it?

event_sale_registration_multi_qty/models/sale_order.py
```diff
@@ -0,0 +1,62 @@
...
+            return super()._init_registrations()
+        Registration = self.env["event.registration"].sudo()
+        for so_line in self.filtered("event_id"):
+            if not so_line.event_id.registration_multi_qty:
+                super(SaleOrderLine, so_line)._init_registrations()
+                continue
+            product_uom_qty = so_line.product_uom_qty
+            # Set temporarily the order line to one avoiding to create more
+            # than one registration
+            so_line.product_uom_qty = 1
```
carlos-lopez-tecnativa, line 48: I'm not sure if this line is related or not, but I attached a video where you can see that the attendance is not created as expected. I expected it to create one attendance with a quantity of 3, but currently it's created with a quantity of 1. Please let me know if I'm correct or if I missed some configuration.


https://github.com/user-attachments/assets/b52134cc-ed77-4413-bf7b-9ea6fc5a2b7b

image

carlos-lopez-tecnativa (changes requested):

event_sale_registration_multi_qty/models/sale_order.py
victoralmau: The problem you describe has been fixed (although the cause was not this part of the code, it was a read-only field in the wizard view).

carlos-lopez-tecnativa (approved):
LGTM

pilarvargas-tecnativa (approved):

PR #474, opened 2025-09-02
[17.0][FIX] event_sale_reservation: Allow product update to register for the event

pilarvargas-tecnativa:
@Tecnativa TT57848

@victoralmau @carlos-lopez-tecnativa please review

### DIFF
--- a/event_sale_reservation/wizards/registration_editor.py
+++ b/event_sale_reservation/wizards/registration_editor.py
@@ -37,6 +37,13 @@ class RegistrationEditor(models.TransientModel):
         "incompatible with each other"."""
         # Modify SO lines to be tickets instead of reservations
         for line in self.event_registration_ids:
+            # HACK: Force product_updatable = True to be able to change the product_id
+            # when converting reservations into registrations.
+            # Overwriting _compute_product_updatable() and controlling this specific
+            # case there is not feasible, given that other modules also overwrite
+            # that compute and change the value, which would result in our logic being
+            # overridden.
+            line.sale_order_line_id.product_updatable = True
             line.sale_order_line_id.write(
                 {
                     "event_id": line.event_id.id,
@@ -45,6 +52,7 @@ class RegistrationEditor(models.TransientModel):
                     "price_unit": line.sale_order_line_id.price_unit,
                 }
             )
+            line.sale_order_line_id._compute_product_updatable()
         # Close current wizard and reopen normally to configure registrations
         upstream_view = self.env.ref("event_sale.view_event_registration_editor_form")
         return {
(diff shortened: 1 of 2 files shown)


event_sale_reservation/wizards/registration_editor.py
```diff
@@ -37,6 +37,7 @@ def action_convert_to_registration(self):
         "incompatible with each other"."""
         # Modify SO lines to be tickets instead of reservations
         for line in self.event_registration_ids:
+            line.sale_order_line_id.product_updatable = True
```
pedrobaeza, line 46: Why not put it in the following `write`?

pilarvargas-tecnativa: Because validation is done before applying the write:
https://github.com/odoo/odoo/blob/17.0/addons/sale/models/sale_order_line.py#L1009-L1011

pedrobaeza: This is definitively not correct. You have to override `_compute_product_updatable` instead. https://github.com/odoo/odoo/blob/17.0/addons/sale/models/sale_order_line.py#L925

pilarvargas-tecnativa: It is not ideal, as there are other modules that inherit that compute and set it to False. I cannot know in what order those computes will be executed to ensure that mine is the last one. 
For example:
https://github.com/odoo/odoo/blob/17.0/addons/sale_project/models/sale_order_line.py#L102

That condition would be met and it would be set to False again.

pedrobaeza: But you can't alter this way a computed field. It's working by chance, as this can be considered a ORM glitch.

Overriding the computed method, call super, and then assign your values. If other overwrites this, maybe a glue module should be developed, or the module overwriting indiscriminately should be fixed.

pilarvargas-tecnativa: I've been reviewing it and it seems that the only option is to create a glue module to resolve this issue. Another option would be to overwrite the write where that restriction is applied using a context flag in the ‘.write’ to decide which logic is executed. I don't know which option you think is more appropriate?

pedrobaeza: Which is the module that breaks this?

pilarvargas-tecnativa: sale_project

https://github.com/odoo/odoo/blob/17.0/addons/sale_project/models/sale_order_line.py#L102

pedrobaeza: OK, seen in a video-call options, and let's go for this hack as the less bad option, adding the proper comment and the test case for having this controlled.

pilarvargas-tecnativa: The error does not appear in the tests because the same product was being used for reservations and registrations. Secondly, creating the order with create is not the most accurate functional test, and thirdly, in the test environment, the order of execution of the computations is not the same as in a production environment. Even so, I have incorporated the most accurate tests so that they can also serve as documentation of the problem.

pedrobaeza (approved):

victoralmau (approved):

PR #480, opened 2025-09-23
[18.0][MIG]Add forbid_duplicates on event_type

wouitmil:


event_registration_partner_unique/models/event.py
```diff
@@ -34,7 +46,7 @@ def _check_forbid_duplicates(self):
             if dupes:
                 # pylint: disable=W8120
                 raise ValidationError(
-                    self.env._("Duplicated partners found in event {0}: {1}.").format(
+                    _("Duplicated partners found in event {0}: {1}.").format(
```
pedrobaeza, line 49: Don't change this.

event_registration_partner_unique/models/event.py
```diff
@@ -51,3 +63,11 @@ def _duplicate_search_domain(self):
             ("attendee_partner_id", "=", self.attendee_partner_id.id),
             ("attendee_partner_id", "!=", False),
         ]
+
+
+class EventType(models.Model):
```
pedrobaeza, line 68: Put this is a new file.

event_registration_partner_unique/views/event_event_view.xml
```diff
@@ -10,7 +10,7 @@
         <field name="priority" eval="99" />
         <field name="inherit_id" ref="event.view_event_form" />
         <field name="arch" type="xml">
-            <xpath expr="//field[@name='create_partner']" position="after">
+            <xpath expr="//field[@name='badge_image']" position="after">
```
pedrobaeza, line 13: Why changing this.

event_registration_partner_unique/models/event.py
```diff
@@ -22,6 +25,15 @@ def _check_forbid_duplicates(self):
             "forbid_duplicates"
         ).registration_ids._check_forbid_duplicates()
 
+    @api.depends("event_type_id")
+    def _compute_forbid_duplicates(self):
+        """Update event configuration from its event type. Depends are set only
```
pedrobaeza, line 30: This is a slight behavior change, as previously changing the type doesn't touch the field, but I think it can be acceptable.

event_registration_partner_unique/__manifest__.py
```diff
@@ -14,6 +14,6 @@
     "license": "AGPL-3",
     "application": False,
     "installable": True,
-    "depends": ["partner_event"],
-    "data": ["views/event_event_view.xml"],
+    "depends": ["event", "partner_event"],
```
pedrobaeza, line 17: No need to add the redundant dependency

PR #491, opened 2026-02-03
[18.0][FIX] event_sale_registration_multi_qty: adapt tests to upstream change

hbrunn:
since https://github.com/OCA/OCB/commit/ba6f21fe884f53e5bcba5aa9bfd376288ab08bf1, event registrations from sale orders aren't autoconfirmed


pedrobaeza (approved):
In fact, I would say that this assert are not relevant to this module, but let's just fix it.

Thanks for the patch.

/ocabot merge patch

PR #498, opened 2026-03-19
[18.0][MIG] website_event_ticket_limit

Highcooley:


mmequignon:
Looks like you squashed pre-commit changes in the translation commit (see readme changes in the second commit).
Those pre-commit changes belongs in their own commit.

Apart from that, code looks good

mmequignon (changes requested):
LG

mmequignon (approved):
wrong button

website_event_ticket_limit/i18n/fr.po
```diff
@@ -0,0 +1,27 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+# 	* website_event_ticket_limit
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Odoo Server 16.0\n"
```
divad1196, line 7: Not critical, but the version number says `16.0` instead of `18.0`.  
Same issue in other traduction files, but this is wrong in many other modules

divad1196 (approved):
All good IMO

meredith-alana (approved):
Max works ✔️ 
Min doesn't, but is in another module and would have to be treated separately. 
Also, not needed here.

PR #525, opened 2026-07-15
[18.0][FIX] website_event_filter_city: remove test key from tour

hbrunn:
the [schema](https://github.com/odoo/odoo/blob/18.0/addons/web_tour/static/src/tour_service/tour_service.js#L46) doesn't contain this key, so this fails on validation


pilarvargas-tecnativa (approved):

pedrobaeza (approved):
Already found this problem and document it the other day:

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0/_compare/6e60c2cdf477d5e602e36efb98f358b7cb914bcd...f0ba8b91b0dbaea38d1f5932f09ba8a566fb64e5

so let's fast-track it:

/ocabot merge patch

PR #530, opened 2026-08-07
[19.0][MIG] website_event_filter_city: Migration to version 19.0

adasatorres-tecnativa:
@Tecnativa TT64025


pedrobaeza:
CI fixed in #531 updating the copier template. I'm rebasing when merged.

website_event_filter_city/controllers/main.py
```diff
@@ -43,26 +43,34 @@ def events(self, page=1, **searches):
...
-            domain = qcontext["countries"][1]["__domain"]
+            domain = qcontext["countries"][1].setdefault("__domain", domain)
             countries_domain = [
                 "|",
                 ("country_id", "=", False),
                 ("country_id", "!=", False),
             ]
             country_tuple_index = next(
-                i for i, x in enumerate(domain) if len(x) > 1 and x[0] == "country_id"
+                (
```
pedrobaeza: Why do you need this parenthesis? Isn't `next` taking the iterator?

website_event_filter_city/static/src/js/website_event_filter_city_tour.esm.js
pedrobaeza, line 1: Why moving this file? And why Git didn't recognize as a move?

website_event_filter_city/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 {
     "name": "Website Event Filter City",
     "summary": "Add a customizable top area to filter events with city",
-    "version": "18.0.1.0.1",
+    "version": "19.0.1.0.1",
```
pedrobaeza: 19.0.1.0.0

pedrobaeza:
CI is failing due to this PR

website_event_filter_city/controllers/main.py
adasatorres-tecnativa: next() is taking the iterator, but the parentheses are required here because the generator expression is not the only argument being passed to next(). Python only allows omitting the generator expression's parentheses when it is the sole argument.

website_event_filter_city/static/src/js/website_event_filter_city_tour.esm.js
adasatorres-tecnativa: I moved it to tests to follow Odoo's recommended folder structure for JS tours. Since this is a tour, I thought it should live under static/src/tests rather than js.

Git didn't recognize it as a move because there were changes to the file as well, so the similarity between the old and new files was probably below Git's rename detection threshold.

pedrobaeza:
When I select a specific city, then the dropdown only shows that city instead all the available ones:

imagen

adasatorres-tecnativa:
> When I select a specific city, then the dropdown only shows that city instead all the available ones:
> 
> imagen

In version 18.0, exactly the same thing happens. I thought it was part of the development design, although it doesn’t really make much sense.

image

From what I’ve seen, it’s because the country is also being used, even though the filter selector comes disabled by default.

pedrobaeza:
Can you please fix it? You can do it in a separate commit.

adasatorres-tecnativa:
@pedrobaeza 
The country filter is disabled in the view where it is implemented. I assume there is some configuration that enables it.
I have adapted the code so that, when the country filter is enabled, selecting a country applies it to the city filter's domain, ensuring that only cities from the selected country are displayed.
When the country filter is not enabled, the city filter continues to work without applying any country restriction.

pedrobaeza:
CI is now red

website_event_filter_city/models/event.py
```diff
@@ -0,0 +1,24 @@
...
+        """Override the original method injecting our city filters"""
+        city = self.env.context.get("event_filter_city")
+        res = super()._search_get_detail(website, order, options)
+        if city:
+            city_domain = [("city", "=", city)]
+            res.update(no_city_domain=res["base_domain"])
+            res["base_domain"].append(city_domain)
+            res["no_country_domain"].append(city_domain)
+            res["no_date_domain"].append(city_domain)
+            self = self.with_context(helloo="Hi there!")
```
carlos-lopez-tecnativa: ????

```suggestion
```

website_event_filter_city/controllers/main.py
```diff
@@ -0,0 +1,82 @@
...
+            (
+                domain
+                for _, name, domain, _ in qcontext["dates"]
+                if name == qcontext["current_date"]
+            ),
+            qcontext["dates"][0][2],
+        )
+        country_filter_view = request.with_user(SUPERUSER_ID).env.ref(
+            "website_event.event_location", raise_if_not_found=False
+        )
```
carlos-lopez-tecnativa: I noticed that the test fails because of this code. Why do you need `with_user(SUPERUSER_ID)`?

You probably need to update the code to use `is_view_active` or the `viewref` function from the website model, because you won't get the correct view based only on the XML ID. You need to take the website into account, as Odoo creates a copy of the views for each website.

https://github.com/odoo/odoo/blob/5b87cfa13a147dfbea2322899f64ad5c1cb837e1/addons/website/models/website.py#L1487-L1503

website_event_filter_city/templates/event.xml
```diff
@@ -0,0 +1,109 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2016-2017 Tecnativa - Jairo Llopis
+     2021 Tecnativa - Víctor Martínez
+     License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl). -->
+<odoo>
+    <template
+        id="event_city"
+        inherit_id="website_event.index_topbar"
+        active="True"
+        customize_show="True"
```
carlos-lopez-tecnativa: You probably need to migrate this to the new website_builder options.

See an example in https://github.com/OCA/vertical-association/tree/19.0/website_membership/static/src/website_builder

More information: https://www.tecnativa.com/my/instances/330/documentation_internal/320

carlos-lopez-tecnativa (changes requested):

carlos-lopez-tecnativa (commented):
LGTM, just a minor question: why is the Cities filter positioned before the Countries filter?

If I filter by a country, only the cities from that selected country are displayed. So, to me, it makes more sense to display the Country filter first and then the Cities filter.

What do you think about this?

adasatorres-tecnativa:
> LGTM, just a minor question: why is the Cities filter positioned before the Countries filter?
> 
> If I filter by a country, only the cities from that selected country are displayed. So, to me, it makes more sense to display the Country filter first and then the Cities filter.
> 
> What do you think about this?

The issue is that when the city selector is added, the country selector is disabled by default. As a result, when you enable it, it is placed at the end. This can be fixed by disabling the city selector and then enabling it again. This way, it should be added at the end.

pedrobaeza:
Can't this be forced with the views priority?

carlos-lopez-tecnativa:
> 
> The issue is that when the city selector is added, the country selector is disabled by default. As a result, when you enable it, it is placed at the end. This can be fixed by disabling the city selector and then enabling it again. This way, it should be added at the end.

I tested this, but it is not working as expected.

I disabled the City filter and enabled it again, but it is still positioned before the Country filter.
I disabled both filters and enabled the Country filter first, followed by the City filter, but the result is the same.

I’ve attached a video showing the issue.


https://github.com/user-attachments/assets/4f4d6113-ecce-40ac-b623-5253c09c3112

adasatorres-tecnativa:
> Can't this be forced with the views priority?

Sure, that's possible. I had forgotten about view priorities. Sorry about that.

carlos-lopez-tecnativa (approved):
LGTM

pedrobaeza (approved):
/ocabot merge nobump

PR #545, opened 2026-08-18
[19.0][FIX] website_event_filter_city: fix visible item count with city filters.

adasatorres-tecnativa:
@Tecnativa TT64025
@pilarvargas-tecnativa @pedrobaeza.

Ensure the city filter takes all other active filters into account when calculating visible events.

### DIFF
--- a/website_event_filter_city/controllers/main.py
+++ b/website_event_filter_city/controllers/main.py
@@ -11,7 +11,7 @@ from odoo.addons.website_event.controllers.main import WebsiteEventController
 
 class WebsiteEvent(WebsiteEventController):
     @route()
-    def events(self, page=1, **searches):
+    def events(self, page=1, slug_tags=None, **searches):
         searches.setdefault("city", self.env._("All Cities"))
         # Inject our city in `_search_with_fuzzy` which ends up in `event.event`
         # `_search_get_detail` override.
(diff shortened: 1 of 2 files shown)


adasatorres-tecnativa:
@pilarvargas-tecnativa, Right now, the city filter should work the same way as the country filter, taking into account that the country filter is affected by the other filters, as well as by the city filter.

dcanete-pyming (commented):
Functional review:

- Verified that selecting a city updates the available countries.
- Verified that selecting a country updates the available cities.
- Verified that text search updates both city and country filters.
- Verified that clearing the filters restores all available options.
- Confirmed that incompatible events are not displayed.

The functional behavior works as expected.

website_event_filter_city/controllers/main.py
```diff
@@ -11,53 +11,43 @@
...
+            "events", search, limit=page * step, order=order, options=options
         )
-        if request.website.is_view_active("website_event.event_location"):
-            country_groups = request.env["event.event"]._read_group(
-                domain, ["country_id"], ["__count"], order="country_id"
+        event_details = details[0]
+        domain_search = (
+            Domain(
+                "name", "ilike", fuzzy_search_term or qcontext["searches"].get("search")
             )
```
pilarvargas-tecnativa: It looks like we're duplicating part of the logic already executed by super().events(). Most of these values are already available in the response qcontext, including the effective search term. Could we avoid duplicating the pagination (step), ordering and _search_with_fuzzy() call, and only retrieve what is actually needed to compute the city counts?

adasatorres-tecnativa: I'm repeating this logic because the search domains are created dynamically in _search_with_fuzzy. They are also not accessible from the parent response context, so I need them for the selector to work correctly.

Also, why should I manually create the domain if there is already a method that does it? I'm not changing any of the returned values; I'm only retrieving the domains I need to correctly build the city selector.

pilarvargas-tecnativa: I see what you mean. The domains themselves are not available directly in qcontext. What I meant is that we shouldn't need to run _search_with_fuzzy() again just to get them.

We can reuse qcontext["searches"] to build the options and get the search details directly:

```
options = self._get_events_search_options(
    slug_tags, **qcontext["searches"]
)
event_details = request.website._search_get_details(
    "events", None, options
)[0]
```

This should give us the domains from event.event._search_get_detail() without having to duplicate the step, ordering, pagination and event search from the parent method.

adasatorres-tecnativa: Sorry, I didn’t realize it could be done this way.

pilarvargas-tecnativa (approved):
Thanks!

carlos-lopez-tecnativa (approved):
LGTM

website_event_filter_city/controllers/main.py
```diff
@@ -11,53 +11,31 @@
...
+    def events(self, page=1, slug_tags=None, **searches):
         searches.setdefault("city", self.env._("All Cities"))
         # Inject our city in `_search_with_fuzzy` which ends up in `event.event`
         # `_search_get_detail` override.
         if searches["city"] != self.env._("All Cities"):
             request.website = request.website.with_context(
                 event_filter_city=searches["city"]
             )
-        response = super().events(page, **searches)
+        response = super().events(page, slug_tags, **searches)
```
pedrobaeza: Don't convert keyword arguments into positional ones, as this breaks other overrides:

```suggestion
        response = super().events(page, slug_tags=slug_tags, **searches)
```

Anyway, it's better to keep the original signature, and if you need anything, just do `searches["slug_tags"]`

pedrobaeza (changes requested):

adasatorres-tecnativa:
ping @pedrobaeza

website_event_filter_city/controllers/main.py
pedrobaeza: You haven't kept the original signature as requested.

adasatorres-tecnativa: Sorry  @pedrobaeza , I may have misunderstood you the other day. Do you mean that I should use the slug_tags from the qcontext in self._get_events_search_options()?

pedrobaeza: No, sorry, I talked about keeping the original signature, but the original signature contains `slug_tags`, so nothing to change. I have reverted to the previous version.

pedrobaeza (approved):
/ocabot merge patch

pedrobaeza:
@adasatorres-tecnativa I already pushed the correct version. It was with `slug_tags`, as the original signature included it. I was confused as at the beginning, it wasn't.

PR #548, opened 2026-08-30
[19.0][IMP] partner_event: prefer the attendee partner as default recipient

moylop260:
The generic default-recipient heuristics of mail pick `partner_id`, which is the person that made the booking, or fall back to the plain email typed in the registration. Mails composed from a registration (mail templates, the composer) are meant for the attendee, so when the module has resolved an `attendee_partner_id` this uses it as the default recipient instead.

When no attendee partner is resolved, the generic heuristics still apply unchanged.

The override keeps the `with_cc`/`all_tos` keyword arguments 19.0 declares on the original and delegates everything else to `super()`.

### DIFF
--- a/partner_event/models/event_registration.py
+++ b/partner_event/models/event_registration.py
@@ -101,6 +101,31 @@ class EventRegistration(models.Model):
         self._update_attendee_partner_id(vals)
         return super().write(vals)
 
+    def _message_get_default_recipients(self, with_cc=False, all_tos=False):
+        """Prefer the attendee partner over the generic heuristics.
+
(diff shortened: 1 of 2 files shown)


moylop260:
@CRogos @nihelgabsi-acsone @pedrobaeza @MohamedOsman7 @vincent-hatakeyama
What do you think?

pedrobaeza (commented):
If there are extra default recipients, this will overwrite them. Can you just replace one by the other?

moylop260:
Good catch!
Thank you!

I have created the changes

Please, review

pedrobaeza (approved):

PR #551, opened 2026-09-16
[19.0][FIX] website_event_filter_city: freeze_time test to avoid event not showing once that date passes.

eduezerouali-tecnativa:
cc @Tecnativa 
ping @pedrobaeza @adasatorres-tecnativa 

Tours were failing due to `date_end`  being hardcoded (2026-09-15), so it no longer appears on /event once that date
passes. Use dates relative to `fields.Datetime.now()` instead.


adasatorres-tecnativa (approved):
LGTM

pedrobaeza:
You may hardcode dates, but if you use `freeze_time`, you will get the same time interval. The problem of variable dates may have problems at midnight due to timezones. Can you assure this is not the case? If problematic, do the `freeze_time` one.

eduezerouali-tecnativa:
> You may hardcode dates, but if you use `freeze_time`, you will get the same time interval. The problem of variable dates may have problems at midnight due to timezones. Can you assure this is not the case? If problematic, do the `freeze_time` one.

I can not assure that will not happen, should not be the case. Anyway I took `freeze_time` path as it looks more safe.
Thanks Pedro.

pedrobaeza:
It seems to not work, as the test is frozen.

pedrobaeza:
So finally is not possible to use `freeze_time`?

eduezerouali-tecnativa:
It pass on local environment, but neither pass OCA CI o our CI. It should be possible to use it as tour goes through but it never terminates the chrome. 

So rollback to first option as is the way odoo manage that case https://github.com/odoo/odoo/blob/47bf5e2e5e91cfbebd30ee83fc92441d551001d3/addons/website_event/tests/test_website_event.py#L59-L66.

pedrobaeza (approved):
OK, let's continue that way.

/ocabot merge patch

